PR #12, opened 2017-08-04
UPDATE for SPI and Wire Lib - shared USIC

mhoechner:
Solves problem for xmc1100 boards that SPI and I2C can not be used at the same time


mhollfelder:
Great, thanks!
I merged the pull request, it will become available with the next release.

Best regards,

Manuel

PR #30, opened 2018-02-19
GPIO modes and tidy up

techpaul:
Adds GPIO pin modes and to keywords etc
    INPUT_PULLDOWN
   OUTPUT_OPENDRAIN

Opendrain outputs one of the most often missing in all Arduino platforms

Make liker.ld for XMC1xxx match XMC4xxxx version adding __stack_start and __stack_end symbols

Tidy up comments in all pins_arduino.h files getting thiongs the same and noting whic pin in pin mapping is ineterrpt 0 or 1


techpaul:
Also minor change to Tone.cpp as one line of code relies on Macro expansion to have a missing bracket.

Makes code more easily readable and matching brackes

PR #42, opened 2018-03-09
Updates for Tone/wiring_time adding RAM totals to compile

techpaul:
Series of commits that

1. Correct Flash total to closer value after compile message
2. Gives A RAM estimate after compile that is at least 16 bytes short due to memory holes not counted, includes all sections even stack which could be 1024 or 2048 bytes depending on XMCLIB version.
3. Rewritten Systick 1ms timer event scheduler that has been tested on 4 Tones and delay (ms) uses hard coded arrays for schedule list and proven methods for scheduling co-operatively, approximate timings is from 5 micro seconds to execute when no tasks waiting and around 20 micro seconds when 5 tasks to be processed. This is on XMC1100. Allows for 4 tones on XMC1 series and 16 on XMC4 series.

Tone had to be modified to use the new scheduler.

Example 4 tones using on board LEDs available on request

4. Add defines for UC_RAM and UC_ALL_RAM for by CPU RAM sizes

### DIFF
--- a/arm/cores/Tone.cpp
+++ b/arm/cores/Tone.cpp
@@ -15,167 +15,142 @@
   License along with this library; if not, write to the Free Software
   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
-
+/* Tones are Systick timer event related not hardware timer related
+   So take care in selecting frequencies range of 1 - 500 Hz supported */
 
 //****************************************************************************
 // @Project Includes
 //****************************************************************************
 #include "Arduino.h"
-#include "pins_arduino.h"
 
 //****************************************************************************
 // @Macros
 //****************************************************************************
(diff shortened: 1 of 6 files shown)


mhollfelder:
Thank you very much for your work here, really amazing!

I merged this for now into the develop branch.
Currently, I am merging all the changes step by step into the release branch together with our changes.

Once I have finished this, I will release the new version.

Best regards,

Manuel

techpaul:
Well been under the weather so a bit behind, need to push up some commits and generate a pull request for a few things

- Fix serial interrupt bug
- Tidy up maths opertaions in serial to simple add/subtract/compare, cuts down interrupt overhead
- Catch up on newer systick method where I had ported code to correct comments and reduce stack variables

Also I will put on my repository some of my test examples for proving multiple tones, accessing new defines etc.

PR #45, opened 2018-03-15
Latest updates tagged V1.06b

techpaul:
These remove TX interrupt being lost, move millis and micro to first possible event, tidy ups for readability
simplify maths functions used on buffer handling for speed, and any size buffer handling.

This will be about my lot for a week or so as dreaded flu is making things doifficult for me.


techpaul:
Latest updates are to match My V1.06c and tidy ups in documentations and unused macros.
This gets a fairly stable version yes things that would be nice to work on but after V1.10 release is my view

mhollfelder:
Great, thank you very much!
I have now merged everything from our side locally and will take your latest changes for V1.1.0.
We are currently working on changing the development process to incorporate such huge feedback from the community directly in development.

I will run some more tests and once everything is done, I will release the new version. 
However, starting from V1.0.1 we should closely align with each other to avoid parallel development.
I will clean the wiki and we should think about additional ways of alignment.

Once again, many thanks and best regards,

Manuel

techpaul:
Good to hear we are getting somewhere.

let me know of any questions I am around a bit at the moment, playing catch up after illness, so many ciustomer jobs to do.

PR #54, opened 2018-04-14
Improve Reset handling and testtone test module

techpaul:
**reset .c .h** Improves reset input thresholds for noise immunity Remove redundant comment at top

**Testtone** uses more LEDs and newer XMC_BOARD contents

Allow  **xmc[1/4]_gpio.h/.c**  to use less restrictions for ability to use --short-enum options on compilers (put potentially smallest values usually enums to end of structure.

**pins_arduino.h** remove more unused defines and variables, get new boards in same style/format as older files, add notes in comments to all variants for Analog inputs that can **only** be digital INPUT (according to datasheet)

Also remove final initialiser element having comma on end that could depending on compiler options create extra empty element


mhollfelder:
Dear Paul,

Sorry for replying so late, currently we have hands full with additional development, but we will prepare a new release to fix the issues for XMC and release new software for our other hardware as well.
Additionally, we are working on all the documentation in parallel.

Thank you very much for all your effort, I will come back to you with updates from our side.

Best regards,

Manuel

techpaul:
Tidy up files and functions
Remove freeRAMStack function show better example with a define macro for get freeStackRAM

Add more comments and functionality to examples

RTC/Sleep to be worked on yet

PR #69, opened 2018-06-27
bug fix

sherylll:
* changed RTC example naming
* removed dependency of deviceControl on DMA


sherylll:
The examples under /libraries are built here: https://github.com/Infineon/platformio-infineonxmc/tree/master/examples

TODO: sync the changes in the 2 repos automatically to avoid having to modify the examples twice.

PR #80, opened 2018-08-03
update CMSIS; add DSP & Radar Lib

sherylll:
Prep for the next release. I2S library to support the examples coming soon.


mphasize:
Hi @sherylll, 

not sure if it's important, but I noticed that the version numbers are still at 1.1.2:
https://github.com/sherylll/XMC-for-Arduino/blob/master/arm/package.json#L5
https://github.com/sherylll/XMC-for-Arduino/blob/master/arm/platform.txt#L8

Cheers,
Marcus

sherylll:
Hey thanks a lot! Just fixed them!

sherylll:
Hi @techpaul 
thanks for pointing out the issues! I have removed all the new operators from the radar lib, however for the FFT/MFCC for XMC4700 there are still many remaining.
Despite being suboptimal, I think we should allow the use of "new" otherwise people might write some code that compiles but does not run and couldn't find out why.
The current workaround is to use a dummy caller:
https://github.com/sherylll/XMC-for-Arduino/blob/9bdf021cb786b4a49a897ba81548895f92d2b4ed/arm/cores/Main.cpp#L30
However a cleaner solution is needed.

techpaul:
Yes @sherylll, we shouild allow new and someone needs to get a cleaner solution into  gcc tree or add that code to the syscalls.h/.c in XMCLIB and CMSIS branch in the right place.

People should be allowed to use 'new' or even malloc, the problem is from experience in embedded those using 'new' tend to be desktop/mainframe programmers or I am learning C++ so the language is right and all machines are the same, or more normally blindly copied a bad example from elsewhere, trying to do things done simpler other ways in embedded systems or at kernel level.

Generally more problems in embedded are caused by new and malloc than they cure.

I am not saying it should not be used but beware of **ALL** the issues using it.

Having seen many experienced people get confused with objects and not know what *'this.'* is at any place in the code, or worse still doing lots of new in recursive calls. Evevn seen Exam board Academics set programming questions with things wrong where they did not know whether the object in a function was
* the object itself
* pointer to the object
* copy of the object
* pointer to the copy of the object
* pointer to the pointer to.........

techpaul:
I note you use a lot of conditionals in the form of
~~~
#if  (UC_FAMILY == XMC1)
~~~

Generally I am using on new library versions define for peripheral, type and qty in **EACH APPLICABLE** pins_arduino.h like
~~~
#define HAS_RTC
~~~
Then using
~~~
#ifndef HAS_RTC
~~~

Easier to maintain and allow for new series or model configurations and updating of less files in the long run.

I would suggest something like XMC4700 pins_arduino.h (and XMC4800 for future proofing) have something like
~~~
#define XMC_USB_4000
~~~
As no doubt any new series will have a different type of controller.

Side note sorting out sbrk blob into HeapMemoryMeasurement to push later and hopefully stop any checks failing.

sherylll:
Hi @techpaul thanks that's a good point.

And what do you mean by sorting out sbrk? Could you mark it in the code?

sherylll:
One build is failing https://travis-ci.org/Infineon/XMC-for-Arduino/jobs/413441572 due to the change in the last merged PR:

https://github.com/Infineon/XMC-for-Arduino/blob/f0aedb5bf0c6c7f56e8672e62802696cb66450e9/arm/variants/XMC1300/config/XMC1300_Sense2GoL/pins_arduino.h#L39

Could you check it?

techpaul:
You are using the wrong define, due to earlier changes in how defines were done at top level boards.txt meant without XMC_BOARD defined as it was the boards.txt build parameters would mean XMC_BOARD would **ALWAYS** be '1' whatever board was in use.

Note the coments above XMC_BOARD as this is for stringifying for human readable format at EXECUTION time on serial/LCD/files. (see example testdefines)
~~~
// XMC_BOARD for stringifying into serial or other text outputs/logs
// Note the actual name XMC and number MUST have a character between 
// to avoid issues with other defined macros e.g. XMC1100
#define XMC_BOARD XMC 1300 Sense2GoL
~~~
The value XMC1300_Sense2GoL is set in boards .txt on (my version) line 171 as
~~~
XMC1300_Sense2GoL.build.board_variant=XMC1300_Sense2GoL
~~~
This is used by build environment from platform.txt in line 71 as
~~~
recipe.cpp.o.pattern="{compiler.path}{compiler.cpp.cmd}" {compiler.cpp.flags} -DXMC{build.board.version}_{build.board.type} -D{build.board_variant} -DF_CPU={build.f_cpu} ......
~~~
Which creates a command line for compilation of
~~~
tools\arm-none-eabi-gcc\5.4-2016q3/bin/arm-none-eabi-g++" -mcpu=cortex-m0 -mthumb -c -g -Os -w -std=gnu++11 -ffunction-sections -fdata-sections -fno-threadsafe-statics -nostdlib --param max-inline-insns-single=500 -fno-rtti -fno-exceptions  -w -x c++ -E -CC -DXMC1302_T038x0064 -DXMC1300_Sense2GoL -DF_CPU=32000000L .....
~~~ 
You need to change in the Radar or other modules
~~~
#if (XMC_BOARD == XMC1300_Sense2GoL)
~~~
To 
~~~
#ifdef XMC1300_Sense2GoL
~~~
Chnaging XMC_BOARD to XMC1300_Sense2GoL in pins_arduino.h (and similar for other boards) will mean it would compile for **ANY** board as would ALWAYS be true and do the OPPOSITE effect to what you want.

sherylll:
The current radar library can also compile for other boards. It is possible to connect the IQ pins of Sense2GoL to other boards and compile the program there.

PR #94, opened 2019-04-04
Changed analog defines A0-A5

Severin9er:
A0-A5 define the pin numbers according to mapping_port_pin[].
analogRead() needs to be changed that it accepts pin numbers as well
This ensures that digitalRead(A0) works.


techpaul:
Please check past closed issues and past updates before suggesting changes and bear in mind there are many boards the changes may have to work on.

On changes like this note that the A0 = 0, method dates back to original AVR Arduino and is the standard method across all platforms

PR #93, opened 2019-03-24
Fixes to XMC2GO pinout, analog functions and PWM

techpaul:
XMC2GO pin 0 and pin 1 were swapped compared to other documentation

Analog functions for init and read improved

PWM prescaler bug in CCU8 gave wrong prescaler

Add more safety, code reduction and documentation

PR #95, opened 2019-04-04
analogRead

Severin9er:
Changed analogRead() so that it accepts both: channel numbers and pin numbers.
Also the defines of the analog pins in pins_arduino.h are changed to match their pin numbers in mapping_port_pin[].
Advantages of this change are that digitalRead(A0) works now properly (previously it deliverd the pin state of digital pin D0).
Also there is a difference now between analog and digital pins, which can be used by software to check whether a pin is analog or digital (e.g. if pin number is out of the range of digital pins, it has to be a analog pin).
And the change improves the compatibility to original Arduino, where it's implemented similar.

### DIFF
--- a/arm/cores/wiring_analog.c
+++ b/arm/cores/wiring_analog.c
@@ -111,12 +111,28 @@ void analogReference( uint8_t ulMode )
 analog_reference = ulMode;
 }
 
-// analogRead takes parameter of ADC channel number
+// analogRead takes parameter of ADC channel number or the pin
 // return 0 for invalid channel
 // Would prefer a return of 0xFFFFFFFF to be obvious in code/debug when wrong
 // instead of valid value as it has been done
-uint32_t analogRead( uint8_t channel )
+uint32_t analogRead( uint8_t pin )
 {
+//in case the given argument is a pin it gets translated to a channel number
+uint8_t channel = pin >= A0 ? pin - A0 : pin;
+	
+//XMC4700 is by now the only XMC Board which has more than 6 analog pins.
+//Therefore also if(NUM_ANALOG_INPUTS > 6) would work here, but it might be incompatible with future boards
+#if defined(XMC4700_Relax_Kit)
+if( pin >= A14 )
+{
+	channel = pin - A14 + 14;
+}
+else if( pin >= A6 )
+{
+	channel = pin - A6 + 6;
+}
+#endif
+	
 uint32_t value = 0;
 
 if( channel < NUM_ANALOG_INPUTS )
--- a/arm/cores/wiring_analog.c
+++ b/arm/cores/wiring_analog.c
@@ -120,14 +120,18 @@ uint32_t analogRead( uint8_t pin )
 //in case the given argument is a pin it gets translated to a channel number
 uint8_t channel = pin >= A0 ? pin - A0 : pin;
 	
-//XMC4700 is by now the only XMC Board which has more than 6 analog pins.
-//Therefore also if(NUM_ANALOG_INPUTS > 6) would work here, but it might be incompatible with future boards
-#if defined(XMC4700_Relax_Kit)
+//If there are more than the standard 6 pins
+#if defined(A14)
 if( pin >= A14 )
 {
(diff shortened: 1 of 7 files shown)


techpaul:
Arduino on AVR does NOT work that way

techpaul:
This does not work with pull request #93

Severin9er:
The new commit should fix the problem with pull request #93.
Also I've had a look in the wiring_analog.c of ArduinoCore-avr and there is a mapping between pin and analog channel in the beginning of analogRead(), isn't it? Also the defines of the analog pins in pins_arduino.h are starting with 14, not with 0. 
Is there a difference to my approach?

techpaul:
First of all your fix for pull request #93 does not work as you have NOT understood the analog side just the digital pin side. Most notably your comment that it is a *standard* to have 6 Analog inputs, this is not true, for a start the XMC2GO **only has 2**. and XMC4700 or XMC4800 has many more What is available depends on the chip used on a board and if some pins have been fixed assigned to other functions, like XMC1100 one analog channel is reserved for Reset pin.

This is why we have the define in all pins_arduino.h files
~~~
#define NUM_ANALOG_INPUTS
~~~
Second the AVR version has changed since I looked at their source over a year ago and some fool has addied this stupid mapping with extra code overhead, I can hazard a guess why, which is **different** to XMC boards. Nearly all actual Arduino boards have idents on the boards that say "A0..." or "Analog Inputs" then  0 to 7.

On XMC boards **NONE** of them are marked that way that is why the wiki here has Pin Out Diagrams, all the pins are marked "P0.1", "P2.3" etc.

Your code makes INVALID assumptions that WILL **FAIL**. 

Most notably that on **ALL boards the Pin number index ALWAYS increases by 1 to match the increase in Analog channel number**. This is not true for example look at pins_arduino.h for 
* XMC1100 Boot Kit and you wil find channel numbers A6 and A7 are **reversed** to pin number index order.
* XMC1300 Boot Kit the order of analog channels to pin numbers goes A6, A7, A9, A8, A11, A10
~~~
	/* 6  */ 	{ XMC_GPIO_PORT2, 10 }, // GPIO / A6 ADC Input				P2.10
	/* 7  */ 	{ XMC_GPIO_PORT2, 11 }, // GPIO / A7 ADC Input				P2.11
	/* 8  */ 	{ XMC_GPIO_PORT2, 2 },  // GPIO / A9 ADC Input				P2.2    (INPUT ONLY)
	/* 9  */ 	{ XMC_GPIO_PORT2, 3 },  // GPIO / A8 ADC Input				P2.3    (INPUT ONLY)
	/* 10  */ 	{ XMC_GPIO_PORT2, 0 },  // I2C Clock SCL / A11 ADC Input	P2.0
	/* 11  */ 	{ XMC_GPIO_PORT2, 1 },  // I2C Data  SDA / A10 ADC Input    P2.1
~~~
This could well be true or worse on any newer board. This now means a simple ADC scan of several channels is difficult
~~~
for( chan =  A0; chan <= A11; chan++ )
    readings[ chan ] = analogRead[ chan ];
~~~
Now has to be a copy paste nightmare of **BAD** coding of multiple analogRead and understanding of reversed channels, and swapping values in arrays. One of my recent projects had **8** analog channels.

If you **REALLY REALLY REALLY MUST** do this change, do better testing and understanding of numbers of channels and their mappings. To make it work for **ANY** board you now have to add a mapping array of pin numbers to Analog channels as A0 to An has lost its index capability for something I doubt many users would even think of doing on an XMC board as the pinout diagrams give the pin number as well as analog channel.

What was a generic function now has many board specific manipulations that will be difficult to maintain for any new boards and will have to be constantly updated by pull requests as the board specific nature is NOT in the variants tree.

Severin9er:
You're right, I see the problem now.
I just see that someone could use analogRead(17) or digitalRead(A0), for example if he uses defines for his pins, like I did. 
But I will close this Pull Request as it's really not working with #93 and will make new implementations complicated.

For my project I will just define the pin as 17 and implement the mapping to analog outside of the core.

Thank you very much for your explanations and your effort.

PR #102, opened 2019-09-09
Fixes to H-Bridge2Go Implementation

9Volts9er:
Pins of PWM and DIS were swapped (also wrong labels on PCB):
PWM is on P2.10 = Arduino Pin 11
DIS is on P2.11 = Arduino Pin 10

Fixes to mapping_pin_PWM4:
- Pin 8 was mapped to non-existing index 10 -> changed to index 0
- Added entry for Pin 11 (PWM of H-Bridge) -> mapped to index 1

PR #107, opened 2019-09-13
Improve IFX9201 library

techpaul:
PWM mode fully tested using XMC11002GO needs SPI and further testing on XMC1100 H-Bridge2GO **BEFORE** merging

REQUIRES previous pull requests 102 and 106 at minimum level.
Add more error codes and returns
Improve scaling of duty cycle to match ANY PWM resolution using analogWrite
Change example into separate SPI and PWM examples
Add examples for SPI and PWM with error trapping
Add more comments
Uses LED library for same action on any other board using same device and library

Helps with issues like #102 

Will make it easier for others to use examples


9Volts9er:
I've just tested all the examples and they worked pretty well. Thanks for this great implementation.
Just one minor thing I had to change in order to get it running:

`digitalWrite( m_Disable, HIGH );` -> this switches all the MOSFETs off.
`digitalWrite( m_Disable, LOW );` -> this switches all the MOSFETs on.

So I think this would need to be changed in the IFX9201.cpp inside:

- `IFX9201::begin( )`

- `IFX9201::forwards( uint8_t duty_cycle )`

- `IFX9201::backwards( uint8_t duty_cycle )`

- `IFX9201::stop( )`

Also PWM was working very well. I haven't tested all of the functions by now, but I will try to do this in the next few days.

techpaul:
Thanks for that I will check through the changes you state over next few days as well.

techpaul:
I have examined the begin and movement functions

The Disable pin polarity was wrong, BUT there is an issue foir SPI mode that changing to setting Disable to HIGH in begin() will STOP SPI mode being set. So I have restructured begin() to set pins according to data sheet and NOT set PWM going in SPI mode. The section of the chip datasheet is in section 4.12.2.1 Control Register section and reads

SPI control
0: Control outputs via PWM/DIR inputs
1: Control outputs via SPI
Note: can only be set if DIS=0 and PWM=0 and DIR=0.
Any change of the DIS, PWM or DIR signals will reset
this bit and revert to standard control via PWM/DIR

So SPI mode only works as 100% or 0% to get any other frequency would need continuous SPI command update of alternate commands of CTRL_REG bit SPWM being 1 then 0. Not sure how SPI mode was really working as a PWM input changing should be changing back to PWM mode, unless there is an errata I have missed for the chip datasheet.

Some updated files for SPI due shortly.

techpaul:
Added additional  changes for better configuration of PWM and DIS especially in begin() to ensure SPI mode can be setup and stay in SPI mode

Note SPI mode has NO PWM control and is either 100% or 0% (ON or OFF) for speed.

techpaul:
Unless I hear otherwise by weekend  19th October I will merge these files

9Volts9er:
Sorry, I was a bit busy the last weeks and therefore I've almost forgotten about this topic, I'm sorry.

Testing result now: Everything what I've tested works fine.

I tried all examples, tried to change from PWM to SPI and vice versa during runtime, used different speeds in PWM mode, tried out-of-range-arguments for backwards() and forwards(), tested setting the PWM-Frequency (from -1 to 20001 to see if out-of-range values get detected). I also tried to call "end()" mid operation and the motor stopped as expected.

Then I tested the register functions. There also everything worked as it should.
I only noticed a little typo in "getCRTLReg()", where the 'R' and 'T' are exchanged. I guess it should be named "getCTRLReg()". This would need to be changed in the .h, .cpp and the keywords.txt.

But regarding funcionality everything is fine. Thanks a lot for this implementation.

techpaul:
Thanks for the testing results.

Have corrected the typo for consistency, will be merged shortly

techpaul:
Now merged thanks for others assitance

PR #109, opened 2019-10-22
Travis administrative update

techpaul:
As motor control examples names changed update Travis automated tests to call correct example files when enabled


techpaul:
Administrative update for Travis tests to use correct examples for test compiles when enabled

Issue #99, opened 2019-07-14
very many compiler warnings


mikrocoder:
Hi,

can you fix this?  I expect flawless conditions from a company like Infineon.   :-)

`
C:\Users\Worker\AppData\Local\Arduino15\packages\Infineon\hardware\arm\1.2.1\cores\xmc_lib\CMSIS\NN\Source\FullyConnectedFunctions\arm_fully_connected_q7.c: In function 'arm_fully_connected_q7':

C:\Users\Worker\AppData\Local\Arduino15\packages\Infineon\hardware\arm\1.2.1\cores\xmc_lib\CMSIS\NN\Source\FullyConnectedFunctions\arm_fully_connected_q7.c:73:90: warning: unused parameter 'vec_buffer' [-Wunused-parameter]

                        const uint16_t out_shift, const q7_t * bias, q7_t * pOut, q15_t * vec_buffer)

                                                                                          ^

C:\Users\Worker\AppData\Local\Arduino15\packages\Infineon\hardware\arm\1.2.1\cores\xmc_lib\CMSIS\NN\Source\FullyConnectedFunctions\arm_fully_connected_mat_q7_vec_q15_opt.c: In function 'arm_fully_connected_mat_q7_vec_q15_opt':

C:\Users\Worker\AppData\Local\Arduino15\packages\Infineon\hardware\arm\1.2.1\cores\xmc_lib\CMSIS\NN\Source\FullyConnectedFunctions\arm_fully_connected_mat_q7_vec_q15_opt.c:122:107: warning: unused parameter 'vec_buffer' [-Wunused-parameter]

                                        const uint16_t out_shift, const q7_t * bias, q15_t * pOut, q15_t * vec_buffer)

                                                                                                           ^

C:\Users\Worker\AppData\Local\Arduino15\packages\Infineon\hardware\arm\1.2.1\cores\xmc_lib\CMSIS\NN\Source\ConvolutionFunctions\arm_nn_mat_mult_kernel_q7_q15_reordered.c: In function 'arm_nn_mat_mult_kernel_q7_q15_reordered':

C:\Users\Worker\AppData\Local\Arduino15\packages\Infineon\hardware\arm\1.2.1\cores\xmc_lib\CMSIS\NN\Source\ConvolutionFunctions\arm_nn_mat_mult_kernel_q7_q15_reordered.c:50:64: warning: unused parameter 'pA' [-Wunused-parameter]

[...]

PR #111, opened 2019-12-09
Mohamemu dev

mohamemu:
This is the pull request to integrate the Radar Baseboard XMC4700 to the IFX package.


arm/cores/wiring_analog.h
```diff
@@ -40,47 +40,108 @@ extern "C" {
...
+     *        Values in between these values vary the duty cycle
      *
      * \param pin
-     * \param value
+     * \param value must be 0 to max value for ADC WRITE resolution
+     *
+     * \return 0 = success, -1 = invalid value, -2 = wrong pin
      */
-    extern void analogWrite( uint8_t pin, uint16_t value ) ;
+    extern int16_t analogWrite( uint8_t pin, uint16_t value ) ;
```
jaenrig-ifx: The Arduino Reference Language should be not be modified. API can be extended with new functionalities, but API compatibility with official boards must be preserved.

arm/platform.txt
```diff
@@ -82,7 +83,8 @@ archive_file_path={build.path}/{archive_file}
...
+#recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" -mthumb --specs=nosys.specs --specs=nano.specs -mcpu={build.mcu} {compiler.c.elf.flags} "-T{build.variant.path}/linker_script.ld" "-Wl,-Map,{build.path}/{build.project_name}.map" {compiler.c.elf.extra_flags} -o "{build.path}/{build.project_name}.elf" "-L{build.path}" -Wl,--cref -Wl,--check-sections -Wl,--gc-sections -Wl,--entry=Reset_Handler -Wl,--unresolved-symbols=report-all -Wl,--warn-common -Wl,--warn-section-align {object_files} "{build.path}/{archive_file}" -lm 
+recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" -mthumb --specs=nosys.specs --specs=nano.specs -mcpu={build.mcu} {compiler.c.elf.flags} "-T{build.variant.path}/linker_script.ld" "-Wl,-Map,{build.path}/{build.project_name}.map" {compiler.c.elf.extra_flags} -o "{build.path}/{build.project_name}.elf" "-L{build.path}" -Wl,--cref -Wl,--check-sections -Wl,--gc-sections -Wl,--entry=Reset_Handler -Wl,--unresolved-symbols=report-all -Wl,--warn-common -Wl,--warn-section-align {object_files} {compiler.libraries.ldflags} "{build.path}/{archive_file}" -lm
```
jaenrig-ifx: There is no extra linker flag after all? 
Was this change necessary?

arm/cores/wiring_analog.c
```diff
@@ -186,66 +241,87 @@ return -1;
 }
 
 
-void analogWrite( uint8_t pin, uint16_t value )
```
jaenrig-ifx: Arduino Reference Language compatibility needs to be ensured. This change breaks its. All API prototypes must be preserved.

arm/cores/wiring_analog.c
```diff
@@ -68,41 +88,76 @@ XMC_VADC_GROUP_SetPowerMode( VADC_G0, XMC_VADC_GROUP_POWERMODE_NORMAL );
...
+
+    Default (minimum) is 8 bits (range from 0 to 255).
+    Maximum is 16 bits (range 0 to 65535)
+
+PC Sept-2019 Change flow to trap invalid input first and leave at old setting if
+    Invalid and return valid value or 255 for error, else return setting used
+     
+    Add set maximum value for resolution
+  */
+uint8_t analogWriteResolution( uint8_t res )
```
jaenrig-ifx: Arduino Reference Language compatibility needs to be ensured. This change breaks its. All API prototypes must be preserved.

arm/cores/Arduino.h
```diff
@@ -85,6 +86,17 @@ extern "C" {
     /*
     * XMC pin interrupt type
     */
+#if defined (XMC4700_Radar_Baseboard)
```
jaenrig-ifx: Conditional compiling for each of the board variants at this level should be avoided. Core components implementation should be consistant and available for all microcontroller families, as long as they have an ERU hardware peripheral. 
It should be generic for all hardware platforms...

arm/cores/HardwareSerial.cpp
```diff
@@ -211,9 +211,13 @@ while( XMC_USIC_CH_GetTransmitBufferStatus( _XMC_UART_config->channel) == XMC_US
 size_t HardwareSerial::write( const uint8_t uc_data )
 {
 // Is the hardware currently busy?
+#if defined (XMC4700_Radar_Baseboard)
```
jaenrig-ifx: Conditional compiling for each of the board variants at this level should be avoided.

arm/cores/WInterrupts.c
```diff
@@ -48,12 +65,55 @@ void CCU40_1_IRQHandler(void)
...
     }
 }
+#endif
 
 void attachInterrupt(uint32_t interrupt_num, interrupt_cb_t callback, uint32_t mode)
 {
     if (interrupt_num < NUM_INTERRUPT)
     {
         XMC_PIN_INTERRUPT_t pin_irq = mapping_interrupt[interrupt_num];
+#if defined (XMC4700_Radar_Baseboard)
```
jaenrig-ifx: Same. Is this specific for the Radar board?

arm/platform.txt
techpaul: There are issues with compilation flags see issue #99 about lots of compiler warnings, many parts in recent compiler update have not had correct machine settings set so boards without FPU have lots of warnings for DSP and FPU XMC_LIB functions which are not used.

The.a library is an Arduino IDE **SESSION** build semi-static library (the make process is not properly adhered to. So everytime you launch the IDE **EVERYTHING** is rebuilt this is the Arduino standard way

arm/cores/wiring_analog.c
techpaul: The Arduino Refernece is a moving target and this does **NOT** break anything as  the ONLY situations this could break anything are NOT supported by Arduino IDE
- Modules compiled in different language (e.g. Forth, Ruby, Java)) and the stubs/wrappers referenced a different version.
- You were able to link in a precompiled binary blob, as the licensing and way of building is everything this method cannot be used

Please remember that 99.999% of users do not realise that member functions like write and Print return a value and ignore it. Alo 99.9% do not know how to use return error codes, this gives even for support the ability to get

Many more users like to have error returns to work out why a call did not work.

If the users does not use the returned value it breaks nothing.

There are many things in the Arduino Reference that are reminiscent of a 90's thesis student project that assumes everything alweys works. Error returns HELP users

You are making lots of assumptions based on ONE board like I have seen before see #95 about assumptions all boards have analog pins in ascending order for port pins

arm/cores/wiring_analog.c
techpaul: See previous comment this will not break anything as the situations where this could even **REMOTELY** break anything are not possible with Arduino licensing

arm/cores/Arduino.h
techpaul: You have that the WRONG way round the core functions should as MUCH as possible be board agnostic and rely on flags for support type like CCU_4 and CCU_8, not board specific (platform) parameters these should be dealt with at LOWER levels

You are making assumptions about your board is the only one being used and the long term support issues it will cause see comments in #95 on a similar vein

arm/cores/HardwareSerial.cpp
techpaul: Wrong way rouns as that is how Arduino Refernces are done for all of their boards with minimal in core level

arm/cores/WInterrupts.c
techpaul: Doing core level changes for specific boards cause LONG TERM support issues use flags for types of interrupt and the board specific file should be used to set appropriate flags

techpaul:
This suggested fix appears to be beased upon a single board and incomplete understanding of how Arduino IDE works.

I vote should be reconsidered and ensure core level are board agnostic

arm/cores/wiring_analog.c
jaenrig-ifx: Understood and agree on the advantages of return codes. Nevertheless, I would not open the door to changes in the Arduino Reference Language core functions, and ensure that sketches are building and running for any compatible platforms. Sketches using the return value of such will not compile for other board families.

techpaul: Whilst I appreciate yor concern many things do not compile acroos Arduino's boards as many are incompatible. Very few boards from Arduino or other makers are compatible .

Until about 18 months ago there was many debates even on Arduino developers m,ailing list about the Arduino Reference stating that **_char_** was unsigned and many examples assumed it was AVR standard of **unsigned** and many SAM/ARM etc boards where **_char_** is **signed**. More recently they added the **_byte_**  datatype and updated the reference

The ways of defining an INPUT_PULLUP varies across Arduino's own boards on AVR you do digitalWrite before PinMode to achieve this all others support INPUT_PULLUP. Other input types vary across boards even in same family.

Let alone the variations on attachinterrupt as the hardware is not compatible or easily able to be one or some cases multiple combinations of types like

- Interrupt
- Change
- High
- Low
- Rising
- Falling

There are other inconsistencies around software PWM, interrupt pins and other features like implementation of SPI.

Whilst your goal is noble I think it is restrictive unnecessarily on a moving target

techpaul: Further proof that others use return codes that are errors see
https://www.espruino.com/EspruinoESP8266#analogread-implementation

Espruino ESP8266 popular wifi unit Arduino compatability
~~~
The esp8266 ADC function is available on pin A0 or without a pin. Using a digital pin will return NaN

Possible ways to call ananlogRead() and the returns:

analogRead(A0);
=0.0029296875

analogRead();
=0.0029296875

analogRead(D0);
=NaN
~~~

jaenrig-ifx: Good and valid point :) In your last example (don´t know the rest) there is a fundamental difference with our xmc-for-arduino: they are pointing out the differences and have documented their implementation details. 

I am an idealist xD and the goal should be to provide a certain level of interoperability, ease-of-use, and the standarization of a common ecosystem. Does it means that the reference language is shared among all the cores and their function prototypes preserved?  Does this just mean that the Arduino cores are just enabled to be used with the Arduino IDE? Or even just a hardware connectivity compatibility? 
These dicussions of 1-to-1 compatibility of built-in libraries and reference language can be endless 👯‍♂ Let´s not make a big deal out of this minor changes (at least for now).

The main issue is to have a board agnostic implementation, try to keep the project maintanable and ensure the features keep working properly with existing and new boards.

techpaul: In an ideal world HardwareSerial would not exist as a core module (and actually could have been achived). Many things are undocumented in the Arduino core like on Serial receive Framing and Overrun errors put grabge data into buffers with no error flags or counts, whilst Pairty Error discards data SILENTLY. Some variations and suggestions have added 8 bit counters to at least count errors.

The XMC-for-Arduino differences to what people **think of as the core reference** are documented not necessarily on this repository or its wiki. I do not remember if I have edit/write permissions to wiki. However for XMC1300 Boot kit on this Wiki points to my Wiki.

For existing differences between _Reference_ and XMC-for-Arduino please see the following pages on my repository Wiki and note some are not changeable to match without MAJOR work

[Functions GPIO](https://github.com/techpaul/XMC-for-Arduino/wiki/Functions-and-Options-Differences)
[Analog functions and improvements](https://github.com/techpaul/XMC-for-Arduino/wiki/Analog-Functions-and-Additions) 

Note some of those changes enable the IFX92101 support to work correctly

PR #119, opened 2020-01-25
Fix simple Bugs in Radar XMC4700

techpaul:
##  pins_arduino.h
Commenting bug on NUM_ANALOG_INPUTS
NUM_PWM did not match list of PWMs assigned

## boards.txt 
Wrong compiler flags that was resolved September 2019 for similar XMC4700 Relax kit
causes redefinition of constants in other modules

Other potential bug exists in
~~~
    const XMC_PIN_INTERRUPT_t mapping_interrupt[] =
~~~
Where this board has 6 entries per row ALL other boards use 5 per row

No time at present to chase this down, I suggest original author @mohamemu  check this

### DIFF
--- a/arm/boards.txt
+++ b/arm/boards.txt
@@ -234,7 +234,10 @@ XMC4700_Radar_Baseboard.build.variant=XMC4700
 XMC4700_Radar_Baseboard.build.board_variant=XMC4700_Radar_Baseboard
 XMC4700_Radar_Baseboard.build.flash_size=2000K
 XMC4700_Radar_Baseboard.build.flash_ld=linker_script.ld
-XMC4700_Radar_Baseboard.build.extra_flags=-DARM_MATH_CM4 -DARM_MATH_DSP -DUSB0 -DINTERRUPT_USE_ERU -DSERIAL_USE_U1C1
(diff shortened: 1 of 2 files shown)


mhollfelder:
Requested review from @mohamemu to test the changes.

mohamemu:
> pins_arduino.h
> Commenting bug on NUM_ANALOG_INPUTS
> NUM_PWM did not match list of PWMs assigned
> boards.txt
> Wrong compiler flags that was resolved September 2019 for similar XMC4700 Relax kit
> causes redefinition of constants in other modules
> Other potential bug exists in
>     const XMC_PIN_INTERRUPT_t mapping_interrupt[] =
> 
> Where this board has 6 entries per row ALL other boards use 5 per row
> No time at present to chase this down, I suggest original author @mohamemu  check this

Yes, this board has 6 entries for mapping_interrupt[] instead of the usual 5 because its interrupt pins are mapped to the ERU instead of PWM. Which is also the reason the flag "INTERRUPT_USE_ERU" is introduced.

mhollfelder:
Merging the pull request into `develop` branch - rebased the original pull request.

Best,

Manuel

PR #123, opened 2020-02-06
Xmc config

danielstuart14:
**Config: Move definitions to separated header file**

These definitions are used by some core functions (Ex: HardwareSerial,
which uses NUM_SERIAL, but including pins_arduino.h causes build to fail)
and need to be included in those files for proper compilation.

Also, this commit allows xmc_spi_conf.h to be further simplified, as it
also defines NUM_SPI.

-------------------------------------------------------------------------
**HardwareSerial: Fix secondary serial interface**

NUM_SERIAL wasn't defined prior to this commit, and because of that,
using Serial1 would cause compilation to abort with error:
"'Serial1' was not declared in this scope"


techpaul:
I am trying to understand why you have split the files. The remainder of the file pins_arduino.h is still specific to that variant.

If you are trying to extend boards to have 2 serial capability then increase things not split them

On many boards like XMC100 Boot Kit XMC1100 XMC2Go etc had only one serial port defined so NUM_SERIAL was 1, so Serial1 was not not defined so the compiler error was CORRECT for the way the variant was configured.

If you wish to add a second serial port then fine add the EXTRA definitions for Serial1 include the pin mapping and interrupts.

Adding NUM_SPI is good as woould NUM_I2C

Looking over files will take a time.

If you wish to add a second serial port then that is fine

techpaul:
Quick look at Arduino.h reveals you are changing the problem in WRONG place as pins_arduino.h is included there

~~~
#include "Print.h"
#include "HardwareSerial.h"

//****************************************************************************
// @Board Variant Includes
//****************************************************************************
#include <pins_arduino.h>
~~~
So splitting file and then adding the split file to be included after Arduino.h is a waste fo time.

Maybe changing the above lines for pins_arduino.h is PART of the solution, but both rely on class definitions in HardwareSerial.h to be referenced BEFORE pins_arduino.h so the last lines of that file will work with
~~~
HardwareSerial Serial( &XMC_UART_0, &rx_buffer_0, &tx_buffer_0 );
~~~

This does not appear to be fixing a real problem as far as I can see without also adding extra parts to pins_arduino.h for definition of Serial1.

Also adding your xmc_config.h after any #include <Arduino.h> is a waste of time as it would be included from pins_arduino.h included from Arduino.h anyway

techpaul:
More checking on files shows

1.  NUM_SERIAL has always be defined
2.  Changes just splits existing variant files in way that will not always work for every variant
3.  Despite claims I see no addition of NUM_SPI

I think there is a base misunderstanding that Arduino first serial port is always referenced as
Serial or Serial0

Any ADDITIONAL serial ports if supported by hardware and variants configuration are named Serial1, Serial2..... So if board does not support Serial1 then referencing it WILL CORRECTLY error

danielstuart14:
Indeed after some thoughts this isn't the best way to solve the issue, but still, without it, Serial1 doesn't work on supported boards (Ex: xmc4700 relax kit). I'll check how to fix that without separating the files

Issue #23, opened 2018-01-11
Tx and Rx pin for UART in XMC 2GO


webstone13:
As shown in the attached image:
![xmc 2go_po](https://user-images.githubusercontent.com/35258248/34809488-4cdd36a4-f6bb-11e7-9cde-2d87fc2e1f70.jpg)

 the UART Rx and Tx pin mentioned are P2.6 and P2.0 respectively. However, in the user manual released by Infineon, the UART Rx and Tx pins mentioned are P2.1 and P2.2. Please find the user manual attached herewith (on page 11 of 15):

[Board_Users_Manual_XMC_2Go_Kit_with_XMC1100_R1.0.pdf](https://github.com/Infineon/XMC-for-Arduino/files/1621312/Board_Users_Manual_XMC_2Go_Kit_with_XMC1100_R1.0.pdf)

Which one is correct? I think the one in the user manual is the right one.

mhoechner:
Hi,

actually both are correct. 
For the XMC it is possible to assign the UART Rx and Tx to different Pins.

In case of this implementation, by default Serial (UART) uses pin P2.1 and P2.2, which enables the connection to the host PC. (Serial Monitor)
If you want to use the on-board Rx and Tx Pins (P2.6 and P2.0), you have to change the "pins_arduino.h" file under:
> XMC-for-Arduino / arm / variants / XMC1100 / config / XMC1100_XMC2GO / 
In this file you have to uncomment line 186 and comment line 185, which will change the configuration of the UART Pins.

Be Aware, that you can only use either the on-board Pins **or** the Pins connected to the Host PC.

BR

mhollfelder:
Actually, that's something we need to document properly and I will follow up on this immediately.

I know, this is confusing at first glance, but due to the structure of the peripherals of the XMC µC series it gets clearer when having a look in the Datasheet/Reference Manual, e.g. [Datasheet XMC1100](https://www.infineon.com/dgdl/Infineon-xmc1100_AB-DS-v01_06-EN.pdf?fileId=5546d4624a0bf290014a4bdaff9325bd) and [Reference Manual XMC1100](https://www.infineon.com/dgdl/Infineon-xmc1100-AA_rm-UM-v01_01-EN.pdf?fileId=5546d46255dd933d0155e31753b077af). 

But even still, there are lots of pages to read for the full picture and it's clearly out of scope of an Arduino-like approach. So, please don't hesitate to approach us for further questions regarding this.

Best regards,

[...]

PR #125, opened 2020-02-06
Create new serial config file

danielstuart14:
This commit separates the serial configuration from pins_arduino.
This is done so pins_arduino can be included earlier on Arduino.h,
and by doing that, Serial1 is fixed on supported boards.


techpaul:
Does this actually work???
Have you tested it as I do not see  this helps.

I suspect the problem is actually something a LOT SIMPLER please check #23 as that has an implementation that was done for XMC1100 XMC2GO for 2 serial and a test sketch I suggest you use to test your fixes.

I suspect the problem is simply the last line of following
~~~
XMC_UART_t XMC_UART_0 =
  {
  .channel              = XMC_UART0_CH0,
  .rx                   = { .port = (XMC_GPIO_PORT_t*)PORT1_BASE,
                            .pin  = (uint8_t)4
                          },
  .rx_config            = { .mode = XMC_GPIO_MODE_INPUT_TRISTATE,
                            .output_level     = XMC_GPIO_OUTPUT_LEVEL_HIGH,
                            .output_strength  = XMC_GPIO_OUTPUT_STRENGTH_STRONG_SOFT_EDGE
                          },
  .tx                   = { .port = (XMC_GPIO_PORT_t*)PORT1_BASE,
                            .pin  = (uint8_t)5
                          },
  .tx_config            = { .mode = (XMC_GPIO_MODE_t) XMC_GPIO_MODE_OUTPUT_PUSH_PULL_ALT2,
                            .output_level     = XMC_GPIO_OUTPUT_LEVEL_HIGH,
                            .output_strength  = XMC_GPIO_OUTPUT_STRENGTH_STRONG_SOFT_EDGE
                          },
  .input_source_dx0     = (XMC_USIC_INPUT_t)USIC0_C0_DX0_P1_4,
  .input_source_dx1     = XMC_INPUT_INVALID,
  .input_source_dx2     = XMC_INPUT_INVALID,
  .input_source_dx3     = XMC_INPUT_INVALID,
  .irq_num              = USIC0_0_IRQn,
  .irq_service_request  = 0
  };

XMC_UART_t XMC_UART_1 =
  {
  .channel              = XMC_UART1_CH0,
  .rx                   = { .port = (XMC_GPIO_PORT_t*)PORT2_BASE,
                            .pin  = (uint8_t)15
                          },
  .rx_config            = { .mode = XMC_GPIO_MODE_INPUT_TRISTATE,
                            .output_level     = XMC_GPIO_OUTPUT_LEVEL_HIGH,
                            .output_strength  = XMC_GPIO_OUTPUT_STRENGTH_STRONG_SOFT_EDGE
                          },
  .tx                   = { .port = (XMC_GPIO_PORT_t*)PORT2_BASE,
                            .pin  = (uint8_t)14
                          },
  .tx_config            = { .mode = (XMC_GPIO_MODE_t) XMC_GPIO_MODE_OUTPUT_PUSH_PULL_ALT2,
                            .output_level     = XMC_GPIO_OUTPUT_LEVEL_HIGH,
                            .output_strength  = XMC_GPIO_OUTPUT_STRENGTH_STRONG_SOFT_EDGE
                          },
  .input_source_dx0     = (XMC_USIC_INPUT_t)USIC1_C0_DX0_P2_15,
  .input_source_dx1     = XMC_INPUT_INVALID,
  .input_source_dx2     = XMC_INPUT_INVALID,
  .input_source_dx3     = XMC_INPUT_INVALID,
  .irq_num              = USIC1_0_IRQn,
  .irq_service_request  = 0
   };
~~~
I suspect the second port's  definition of irq_service_request should be 1 as in
~~~
  .irq_service_request  = 1
~~~

danielstuart14:
Yes it does, "HardwareSerial.h" checks NUM_SERIAL for the amount of serial ports available, but, as seeing on "Android.h", the "pins_arduino.h" is only included after "HardwareSerial.h".
This way, even on xmc4700 relax kit (which has NUM_SERIAL = 2), the second serial was never enabled/defined, see this line: https://github.com/Infineon/XMC-for-Arduino/blob/0fa47ed19a65ace89f021b51b33c64ef414607cb/arm/cores/HardwareSerial.h#L130
That's what I was talking about when saying that NUM_SERIAL wasn't defined.

This commit only separates pins_arduino into two files because all the serial related definitions need to be done after the include of "HardwareSerial.h", but, without defining NUM_SERIAL before "HardwareSerial.h" include the second serial would never work.

Also, this commit doesn't add/remove any serial interface. It just make the secondary serial to work on boards that already have it enabled.

You can check it on boards with two Serials (Ex: xmc4700 relax kit), before this commit, Serial1 wasn't even declared.

techpaul:
I have delved into the core files and there are a series of issues and I think simpler solutions

1.  HardwareSerial.h and .cpp are NOT board agnostic by use of NUM_SERIAL
2.  HardwareSerial.cpp includes HardwareSerial.h which early on includes Arduino.h which again includes HardwareSerial.h all before pins_arduino.h

Suggested and more maintainable solution

- Move board specific irq handlers event handlers and extern definitions to pins_arduino.h for each variation in different sections for ARDUINO_MAIN and always sections
- Change HardwareSerial.cpp to include Arduino.h which will include HardwareSerial.h
- Remove include Arduino.h from HardwareSerial.h 

With SOME of the above done and doing the following at end of XMC4700 Relax Kit/pins_arduino.h before last #endif
~~~
#ifdef __cplusplus
extern HardwareSerial Serial;
extern HardwareSerial Serial1;
#endif
~~~

Solved the definition problem for me

techpaul:
Some of this issue is probably the build operation not using cpp build instructions for a .ino file

@mhollfelder Probably want to check your latest build mods 

I will sort out a simpler set of fixes making HardwareSerial less board specific as classes should be

AlucardMode0:
> I have delved into the core files and there are a series of issues and I think simpler solutions
> 
> 1. HardwareSerial.h and .cpp are NOT board agnostic by use of NUM_SERIAL
> 2. HardwareSerial.cpp includes HardwareSerial.h which early on includes Arduino.h which again includes HardwareSerial.h all before pins_arduino.h
> 
> Suggested and more maintainable solution
> 
> * Move board specific irq handlers event handlers and extern definitions to pins_arduino.h for each variation in different sections for ARDUINO_MAIN and always sections
> * Change HardwareSerial.cpp to include Arduino.h which will include HardwareSerial.h
> * Remove include Arduino.h from HardwareSerial.h
> 
> With SOME of the above done and doing the following at end of XMC4700 Relax Kit/pins_arduino.h before last #endif
> 
> ```
> #ifdef __cplusplus
> extern HardwareSerial Serial;
> extern HardwareSerial Serial1;
> #endif
> ```
> 
> Solved the definition problem for me

soo what should i do to make Serial1 work ?

techpaul:
Use the develop branch for now until a new version is released, as that incorporates fixes detailed above and in the merged files to ensure these things happen correctly.

See pull request #126

PR #129, opened 2020-02-12
Config: Set correct PCLK values

danielstuart14:
These boards had their PCLK set to a wrong value, causing the function
"setAnalogWriteFrequency" to misbehave.
An example can be seen in the image below, where a frequency of 50hz was set
on a XMC4700 Relax board:
 - Previous: https://i.imgur.com/9vfwv1d.jpg
 - After: https://i.imgur.com/YGaZxsM.jpg

PCLK values were taken from the datasheets of each MCU. For XMC1400, the PCLK
also matches the one set by its initialization/startup rotine.

### DIFF
--- a/arm/variants/XMC1400/config/XMC1400_Boot_Kit/pins_arduino.h
+++ b/arm/variants/XMC1400/config/XMC1400_Boot_Kit/pins_arduino.h
@@ -52,7 +52,7 @@
 
 #define PWM4_TIMER_PERIOD (2041U)  // Generate 490Hz @fCCU=1MHz
 
-#define PCLK 48000000u 
+#define PCLK 96000000u 
  
 #define PIN_SPI_SS    10
 #define PIN_SPI_MOSI  11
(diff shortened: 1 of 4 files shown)


techpaul:
Well done when I was looking at other issues in pins_arduino.h for various boards the 4700 and 4800 at least just looked wrong for PCLK.

It was a back burner issue to get back to for me working my way through libraries and sorting other issues at moment between other work.

danielstuart14:
@jaenrig-ifx I tested it only on xmc4700 Relax, as it's the only board affected that I have on hands.
I would be very glad if someone tested this patch with all boards affected, but it should be working.

Both xmc4700 Radar and xmc4800 relax have the same pclk as xmc4700 relax, they use the same oscillator. This can be confirmed by checking their user manuals:
https://www.infineon.com/dgdl/Infineon-Board_User_Manual_XMC4700_XMC4800_Relax_Kit_Series-UM-v01_02-EN.pdf?fileId=5546d46250cc1fdf01513f8e052d07fc
https://www.infineon.com/dgdl/Infineon-AN602%20Radar%20Baseboard%20XMC4700-ApplicationNotes-v01_00-EN.pdf?fileId=5546d4626e651a41016e82b63d8a1574

Also, xmc1400 sets the pclk to double the mclk, which is set to 48mhz:
https://github.com/Infineon/XMC-for-Arduino/blob/master/arm/variants/XMC1400/system_XMC1400.c

jaenrig-ifx:
Perfect :) Okay. Definitively, at least now the datasheets nominal values are updated.  I merge it.
Thanks @danielstuart14 !

PR #131, opened 2020-02-17
Updates on XMC4700 Radar baseboard

mkiessling-ifx:
* Fixed issue with missing "compiler.libraries.ldflags", if Radar_S2GL_Pulsed_Doppler_lib is not used.

* adding lib for "arm"
* fixed names and callback types
* Changed board name in README.md
* Added 2nd UART for XMC Radar baseboard on pin P9.3 and P9.4 (= Connector P2, (10 pin header), Pin 4 and 5) and tested both serial ports for TX and RX


techpaul:
Once conflicts are resolved and the Assets images (or images moved into same repository) this tidies up many things noticed missing in previous update

techpaul:
Note in library.properties you are WRONGLY stating

    architectures=cortex-m4,arm

In library structure the "arm" part refers to  library paths in our case

             .../.rduino15/packages/Infineon/hardware/arm

Which then has version number appended for directory search patterns, we do not have a cortex-m4 path there.

The only place "coretx-m4" is needed for is when the precompiled flag is set and the folder

             .../.rduino15/packages/Infineon/hardware/arm/<ver>/libraries/<name>/src/{build-mcu}

is added to library serach path, and {build.mcu} is defined in boards.txt
~~~
XMC4700_Radar_Baseboard.build.mcu=cortex-m4
~~~

See https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification

If it does not work without your settings there is a bug in Arduino IDE

techpaul:
Is this going to be fixed or considered a closed test

mkiessling-ifx:
I removed arm lib variant in my latest commit. 
Thank you for reviewing!

techpaul:
Removed the WRONG lib variant as you do not UNDERSTAND what the architecture statement refers to or how it works.

YOUR patches are NOT approved by me and are a mess.

PR #132, opened 2020-02-18
Tidy up XMC_BOARD and XMC_BOARD_NAME and platform.txt

techpaul:
## History
XMC4700 and XMC4700 Relax kit used as examples for ALL variants

Over past revisions of the board package, mainly Arduino changes, build changes, XMCLIB changes some definitions have changed, most notably definitions of processor type and naming.

At one time for example XMC4700 was not defined, later it became defined in 

-  platform.txt  "build.variant" which got passed to compiler as -D\{build.variant\} e.g. -DXMC4700
-  XMCLib - xmc_device.h defined this as (4700)

So utilising anything with XMC4700 especially in stringifying (working before changes) or adding quotes made strange output strings like 

-  1
-  (4700) Relax Kit
-  "XMC4700 Relax Kit"
-  \"\"XMC4700 Relax Kit\"\"

So XMC_BOARD rules became specify XMC4700 as XMC 4700 to avoid issues

This was primarily meant for multi-board libraries to output which board was being used in normal human readable format matching what they see as board name on packaging, then shown on serial port or to LCD displays or similar for prettifying, confidence checks and help in assisting issue solving (like running on non-supported board).

Human Readable Format is so that even managers shown the outputs can understand them, not just developers well versed in the lower levels of XMC-for-Arduino, XMCLib and CMSIS

At one stage in the testing XMC_BOARD changed to XMC_BOARD_NAME to find what was causing issues, unfortunately when it was reverted to XMC_BOARD not all associated usages were changed back as well or suitably changed to use the correct constants.

## Changes
Looking back through files and rationalising all XMC_BOARD_NAME to XMC_BOARD and NOT using those for selective compiling in #if statements, but use better UC_FAMILY, UC_DEVICE etc or check if specific \{build.board_variant\} is defined e.g.
~~~
#if defined( XMC4700_Relax_Kit )
~~~
Do **NOT** use \{build.variant\} definitions in any #if statement of code as they are **ALL** defined in xmc_device.h so all the following are **ALWAYS** defined
~~~
#define    XMC4800    (4800)
#define    XMC4700    (4700)
#define    XMC4500    (4500)
#define    XMC4502    (4502)
#define    XMC4504    (4504)
#define    XMC4400    (4400)
#define    XMC4402    (4402)
#define    XMC4300    (4300)
#define    XMC4200    (4200)
#define    XMC4100    (4100)
#define    XMC4104    (4104)
#define    XMC4108    (4108)
#define    XMC1401    (1401)
#define    XMC1402    (1402)
#define    XMC1403    (1403)
#define    XMC1404    (1404)
#define    XMC1300    (1300)
#define    XMC1301    (1301)
#define    XMC1302    (1302)
#define    XMC1200    (1200)
#define    XMC1201    (1201)
#define    XMC1202    (1202)
#define    XMC1203    (1203)
#define    XMC1100    (1100)
~~~
To simplify things and reduce unnecessary compiler flags  -D\{build.variant\} should **NEVER** be used in platform.txt  as serves NO purpose, has been used in the past, but not any longer.

Also simplifies use of linker script in platform.txt to use define \{build.flash_ld_path\} created at top of platform.txt to allow for different linker scripts for different sub-variants from boards.txt to work properly.

Updated files are in folders libraries and tests, If any have been missed please let me know


techpaul:
As lack of comments to contrary merged to develop

PR #137, opened 2020-05-31
Remove unicode apostrophe

pfeerick:
Fixes https://github.com/Infineon/platformio-infineonxmc/issues/12, as this unicode character causes the `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 257: invalid start byte:` error that has been fouling the 1.1.4 builds.


techpaul:
Seems simple enough to me but would prefer in develop branch first ready for next version release

pfeerick:
Repeated in #develop

PR #138, opened 2020-06-01
Remove unicode apostrophe

pfeerick:
Fixes https://github.com/Infineon/platformio-infineonxmc/issues/12, as this unicode character causes the `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 257: invalid start byte:` error that has been fouling the 1.1.4 builds. since April 6th.


pfeerick:
I've repeated the change I made against the master branch. Any timeline as to when the next version is likely to happen? As this has resulted in broken builds for the linked platform since April.

danielstuart14:
yup, development branch should be released as soon as possible. The clock issue with xmc2go is still present on stable.

techpaul:
Much as I would like to release a new version that is beyond my PERSONAL control but it has been merged into develop branch

For those that is a problem for please downlaod develop branch either by zip or git.

pfeerick:
Tsk, Tsk! Well, at least it has been locked in to be included in the next release... can't ask for much more! :) Thanks! 👍

PR #143, opened 2020-06-15
Tone edge case fix

techpaul:
Mainly counter EDGE case of requesting a tone before first Systick interrupt after power up or reset. Setting a variable to normal _MAX_TASKS instead of 0.

Some other minor spelling/tidy ups in Tone to remove redundant assignments.

Tested on XMC4700 Relax Light using develop branch at this date

Will leave for a week for comments and merge


techpaul:
As no comments against this has been merged into develop

PR #154, opened 2020-07-01
Develop - Prepare for Architecture change

techpaul:
Add extra documentation to assist changing from 'arm' to 'xmc' architecture name, and have less hassles over changeover.

Update libraries for now to do both, and later to do just new architecture

Remove Hibernate functionality as this cannot be performed within Arduino framework, especially flag that we have recovered from Hibernate being in NON-volatile storage not possible. No save state for any application or device configurations possible.

Small tidy ups for readability


techpaul:
Usual caveat on this PR that will be left for others to review for a few days before merging

PR #155, opened 2020-07-07
Libraries: Add Servo library

danielstuart14:
This was completely written from scratch, with exception of the example
files, which are from Arduino's repository with some changes.

This is an essential library, as some default examples from the IDE use
the Servo library.

Also, it uses xmc's full potential, generating the PWM by hardware.

### DIFF
--- a/libraries/Servo/examples/Knob/Knob.ino
+++ b/libraries/Servo/examples/Knob/Knob.ino
@@ -13,23 +13,19 @@
 
 #include <Servo.h>
 
-#define RESOLUTION 16 // sets PWM resolution to 16 bits (max value: 2^16 - 1 = 65535)
-
 Servo myservo;  // create servo object to control a servo
 
 const uint8_t potpin = 0;  // analog pin used to connect the potentiometer
-const uint16_t maxAnalog = (1 << RESOLUTION) - 1; //Uses bitshift for efficient compile-time calculation
-
 uint16_t val;    // variable to read the value from the analog pin
 
 void setup() {
-  analogWriteResolution(RESOLUTION); //Needed for better accuracy when driving a Servo
-  myservo.attach(9);  // attaches the servo on pin 9 to the servo object
+    analogWriteResolution(16); // sets PWM resolution to 16 bits, for better servo accuracy
+    myservo.attach(9);  // attaches the servo on pin 9 to the servo object
 }
 
 void loop() {
-  val = analogRead(potpin);            // reads the value of the potentiometer (value between 0 and 2^RESOLUTION - 1)
-  val = map(val, 0, maxAnalog, 0, 180);     // scale it to use it with the servo (value between 0 and 180)
-  myservo.write(val);                  // sets the servo position according to the scaled value
-  delay(15);                           // waits for the servo to get there
+    val = analogRead(potpin); // reads the value of the potentiometer
+    val = map(val, 0, getAnalogReadMaximum(), 0, 180); // scale it to use it with the servo (value between 0 and 180)
+    myservo.write(val); // sets the servo position according to the scaled value
+    delay(15); // waits for the servo to get there
 }
--- a/libraries/Servo/examples/Knob/Knob.ino
+++ b/libraries/Servo/examples/Knob/Knob.ino
@@ -1,0 +1,35 @@
+/*
+ Controlling a servo position using a potentiometer (variable resistor)
+ by Michal Rinott <http://people.interaction-ivrea.it/m.rinott>
+
+ modified on 8 Nov 2013
+ by Scott Fitzgerald
+ http://www.arduino.cc/en/Tutorial/Knob
+
+ modified 7 Jul 2020
+ by Daniel Stuart <daniel.stuart@pucpr.edu.br>
+ Added analogWriteResolution()
+*/
+
+#include <Servo.h>
+
+#define RESOLUTION 16 // sets PWM resolution to 16 bits (max value: 2^16 - 1 = 65535)
+
+Servo myservo;  // create servo object to control a servo
+
+const uint8_t potpin = 0;  // analog pin used to connect the potentiometer
+const uint16_t maxAnalog = (1 << RESOLUTION) - 1; //Uses bitshift for efficient compile-time calculation
+
+uint16_t val;    // variable to read the value from the analog pin
+
+void setup() {
+  analogWriteResolution(RESOLUTION); //Needed for better accuracy when driving a Servo
+  myservo.attach(9);  // attaches the servo on pin 9 to the servo object
+}
+
+void loop() {
+  val = analogRead(potpin);            // reads the value of the potentiometer (value between 0 and 2^RESOLUTION - 1)
+  val = map(val, 0, maxAnalog, 0, 180);     // scale it to use it with the servo (value between 0 and 180)
+  myservo.write(val);                  // sets the servo position according to the scaled value
+  delay(15);                           // waits for the servo to get there
+}
--- a/libraries/Servo/examples/Knob/Knob.ino
+++ b/libraries/Servo/examples/Knob/Knob.ino
@@ -27,5 +27,5 @@ void loop() {
     val = analogRead(potpin); // reads the value of the potentiometer
     val = map(val, 0, getAnalogReadMaximum(), 0, 180); // scale it to use it with the servo (value between 0 and 180)
     myservo.write(val); // sets the servo position according to the scaled value
-    delay(15); // waits for the servo to get there
+    delay(20); // waits for the servo to get there
 }
--- a/libraries/Servo/examples/Sweep/Sweep.ino
+++ b/libraries/Servo/examples/Sweep/Sweep.ino
@@ -1,0 +1,39 @@
+/* Sweep
+ by BARRAGAN <http://barraganstudio.com>
+ This example code is in the public domain.
+
+ modified 8 Nov 2013
+ by Scott Fitzgerald
+ http://www.arduino.cc/en/Tutorial/Sweep
+
+ modified 7 Jul 2020
+ by Daniel Stuart <daniel.stuart@pucpr.edu.br>
+ Added analogWriteResolution()
+*/
+
+#include <Servo.h>
+
+
+#define RESOLUTION 16 // sets PWM resolution to 16 bits (65535 max)
+
+Servo myservo;  // create servo object to control a servo
+// twelve servo objects can be created on most boards
+
+int pos = 0;    // variable to store the servo position
+
+void setup() {
+  analogWriteResolution(RESOLUTION); //Needed for better accuracy when driving a Servo
+  myservo.attach(9);  // attaches the servo on pin 9 to the servo object
+}
+
+void loop() {
+  for (pos = 0; pos <= 180; pos += 1) { // goes from 0 degrees to 180 degrees
+    // in steps of 1 degree
+    myservo.write(pos);              // tell servo to go to position in variable 'pos'
+    delay(15);                       // waits 15ms for the servo to reach the position
+  }
+  for (pos = 180; pos >= 0; pos -= 1) { // goes from 180 degrees to 0 degrees
+    myservo.write(pos);              // tell servo to go to position in variable 'pos'
+    delay(15);                       // waits 15ms for the servo to reach the position
+  }
+}
--- a/libraries/Servo/examples/Sweep/Sweep.ino
+++ b/libraries/Servo/examples/Sweep/Sweep.ino
@@ -26,11 +26,11 @@ void loop() {
     for (pos = 0; pos <= 180; pos += 1) { // goes from 0 degrees to 180 degrees
         // in steps of 1 degree
         myservo.write(pos); // tell servo to go to position in variable 'pos'
-        delay(15); // waits 15ms for the servo to reach the position
+        delay(20); // waits 20ms for the servo to reach the position
     }
     
     for (pos = 180; pos >= 0; pos -= 1) { // goes from 180 degrees to 0 degrees
         myservo.write(pos); // tell servo to go to position in variable 'pos'
-        delay(15); // waits 15ms for the servo to reach the position
+        delay(20); // waits 20ms for the servo to reach the position
     }
 }
--- a/libraries/Servo/examples/Sweep/Sweep.ino
+++ b/libraries/Servo/examples/Sweep/Sweep.ino
@@ -13,27 +13,24 @@
 
 #include <Servo.h>
 
-
-#define RESOLUTION 16 // sets PWM resolution to 16 bits (65535 max)
-
-Servo myservo;  // create servo object to control a servo
+Servo myservo; // create servo object to control a servo
 // twelve servo objects can be created on most boards
 
-int pos = 0;    // variable to store the servo position
-
 void setup() {
-  analogWriteResolution(RESOLUTION); //Needed for better accuracy when driving a Servo
-  myservo.attach(9);  // attaches the servo on pin 9 to the servo object
+    analogWriteResolution(16); // sets PWM resolution to 16 bits, for better servo accuracy
+    myservo.attach(9); // attaches the servo on pin 9 to the servo object
 }
 
 void loop() {
-  for (pos = 0; pos <= 180; pos += 1) { // goes from 0 degrees to 180 degrees
-    // in steps of 1 degree
-    myservo.write(pos);              // tell servo to go to position in variable 'pos'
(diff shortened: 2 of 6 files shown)


techpaul:
There is at lest one issue in examples
~~~
const uint16_t maxAnalog = (1 << RESOLUTION) - 1; //Uses bitshift for efficient compile-time calculation
~~~

Whilst this is true for Write resolution it is not for Read resolution, as they are different. Read has max of 12 bit default 10 bit

wiringAnalog has additions to get this value in functions denoted on anlog section of wik]( https://github.com/Infineon/XMC-for-Arduino/wiki/Analog-Functions-and-Additions)

Namely
1. getAnalogReadMaximum
2. getAnalogWriteMaximum

Specifically added to help other functions that do PWM on percentage to scale properly

danielstuart14:
> There is at lest one issue in examples
> 
> ```
> const uint16_t maxAnalog = (1 << RESOLUTION) - 1; //Uses bitshift for efficient compile-time calculation
> ```
> 
> Whilst this is true for Write resolution it is not for Read resolution, as they are different. Read has max of 12 bit default 10 bit
> 
> wiringAnalog has additions to get this value in functions denoted on anlog section of wik]( https://github.com/Infineon/XMC-for-Arduino/wiki/Analog-Functions-and-Additions)
> 
> Namely
> 
> 1. getAnalogReadMaximum
> 2. getAnalogWriteMaximum
> 
> Specifically added to help other functions that do PWM on percentage to scale properly

You are right, I completely forgot about the read resolution. My bad!
I'll update it using those functions!


I usually like using "this->" even when not needed, as it makes easier to know that it's attribute by just glancing at it. But I'm going to change it.

danielstuart14:
Also, the delay in the loop is so the servo has enough time to change its position before receiving another command.

danielstuart14:
Also, the delay in the loop is so the servo has enough time to get there, before receiving another command.
The sweep example wouldn't work, for example, if it didn't had a delay.

danielstuart14:
About the issue with timing, it's likely that it was because of the wrong crystal frequency that was defined. It should be correct now.

techpaul:
> Also, the delay in the loop is so the servo has enough time to get there, before receiving another command.
> The sweep example wouldn't work, for example, if it didn't had a delay.

Wrong way round the FASTEST RC style Servos should be updated (amongst common specs) is 50Hz i.e. 20ms. This needs be set in the setAnalogWriteFrequency when you attach the servo, and is then FIXED (unlike in AVR)

The AVR style with their overheads of chaining delays on timers without using standard channel timings means like quite a lot of  AVR style libraries it meets the specs under **SOME** conditions. A delay of 15 ms in the examples is TOO FAST when using Hardware PWM means you have a creeping update cyctl where at some point you will get an update happening either during a high pulse or worst case missed or causes jitter.

There are MANY things wrong with the AVR and other standard versions like resolution down to 4 micro seconds, I have used PWM down to 0.5 micro seconds on XMC1100 and XMC1300.

As Hardware PWM repeats the same signal setting UNTIL changed updating at 15 ms intervals defeats the purpose of using PWM for a single channel. If you have
1. delay 100ms (update rate 10Hz) it shows PWM continuing and actually works at human adjustable rates
2. You never get missed or other issues on steps
3. If you sweep on 3 degree steps at 100 ms delay per loop time on a sweep on a SIMPLER sweep algoritm gives 6 seconds 0 to 180 degrees.

Been testing this on examples using 2 and 4 channel outpts.

I have not checked all the code but the AVR method does NOT use min/max settings in attach beyond saving them. They use a range of 544 to 2400 which should be your default extremes. I attach a few doc files as azip that I created for where I was up to so far

[readme.zip](https://github.com/Infineon/XMC-for-Arduino/files/4887406/readme.zip)

techpaul:
> I usually like using "this->" even when not needed, as it makes easier to tell that something is an instance variable by just glancing at it. But I'm going to change it.

This method is ONLY appropriate when passing references/pointers around, by using a class the method automatically has this-> passed in and only on very old C++ compilers was it necessary to show it was explicitly this class method. Scope of not file, local, or global means match with class if present.

I beleive you are copying errors unknown to you) from Arduino's bad code into your version, which is why I was doing some extensive tests.

techpaul:
> 
> 
> About the issue with timing, it's likely that it was because of the wrong crystal frequency that was defined. It should be correct now.

Timing issues appear in many places my tests on timing were for servo output channels where scaling was wrong and NON-linear  600 to 1200 range being changed to 700 to 2000 with a multiply of 2.

I have been testing so far on Boot Kits for XMC1100 and XMC1300, not touched XMC4700 or XMC11002GO or others yet

danielstuart14:
> > Also, the delay in the loop is so the servo has enough time to get there, before receiving another command.
> > The sweep example wouldn't work, for example, if it didn't had a delay.
> 
> Wrong way round the FASTEST RC style Servos should be updated (amongst common specs) is 50Hz i.e. 20ms. This needs be set in the setAnalogWriteFrequency when you attach the servo, and is then FIXED (unlike in AVR)
> 
> The AVR style with their overheads of chaining delays on timers without using standard channel timings means like quite a lot of AVR style libraries it meets the specs under **SOME** conditions. A delay of 15 ms in the examples is TOO FAST when using Hardware PWM means you have a creeping update cyctl where at some point you will get an update happening either during a high pulse or worst case missed or causes jitter.
> 
> There are MANY things wrong with the AVR and other standard versions like resolution down to 4 micro seconds, I have used PWM down to 0.5 micro seconds on XMC1100 and XMC1300.
> 
> As Hardware PWM repeats the same signal setting UNTIL changed updating at 15 ms intervals defeats the purpose of using PWM for a single channel. If you have
> 
> 1. delay 100ms (update rate 10Hz) it shows PWM continuing and actually works at human adjustable rates
> 2. You never get missed or other issues on steps
> 3. If you sweep on 3 degree steps at 100 ms delay per loop time on a sweep on a SIMPLER sweep algoritm gives 6 seconds 0 to 180 degrees.
> 
> Been testing this on examples using 2 and 4 channel outpts.
> 
> I have not checked all the code but the AVR method does NOT use min/max settings in attach beyond saving them. They use a range of 544 to 2400 which should be your default extremes. I attach a few doc files as azip that I created for where I was up to so far
> 
> [readme.zip](https://github.com/Infineon/XMC-for-Arduino/files/4887406/readme.zip)

Maybe I got your question wrong, but, the delay itself isn't for the timing of the pwm signal. We still use a 50hz signal, generated using XMC libraries. The delay is so the motor, as a electromechanical part which has some delay (between getting the signal, processing it, generating induction, etc), can get to the position before our mcu sends another one. It's an arbitrary value, maybe, but that's what the default example code uses, and I was not going to change it to another arbitrary value.

Now, those are examples, the library itself doesn't have any delay of this kind. In those examples we want:
-Sweep: The motor should move back and forth in a smooth stepping. Note that an example should always work for the user, as he may not be that experienced. So, maybe his servo isn't fast enough to change 1 degree in less than 15ms, that's why we use a delay here. 
-Knob: Here we want the servo to move with the value it read from a potentiometer, again, we use a delay so we don't spam the servo with like 200 changes in the PPM before it even got to the end of the 20ms wave (which may happen if the user starts to rotate the knob fast). 

But again, the library itself is fully based on hardware pwm generation, and doesn't have a single delay.

Now, about the "this->" stuff, this is my preference when writting any object oriented language (java, c++, etc). I use it wherever possible as it makes easier for me to read my own code. I change it as I see this isn't how code is done here, no problem. It isn't about "copying errors unknown".

About the 544 to 2400 default values, you are right. This was what I used for my own project, I forgot that I changed that. I'll change it back.

danielstuart14:
Actually, about the default ms values, idk what Arduino does, but the industry default values are between 1000 and 2000ms for a single channel PPM servo.

There are many sources for this, you can check in the web, but here is one with it very detailed: https://dronebotworkshop.com/servo-motors-with-arduino/

Also from wikipedia: https://en.wikipedia.org/wiki/Servo_control#/media/File:Sinais_controle_servomotor.JPG

techpaul:
**Please remember many parts of Standard Arduino code, libraries and examples has many historical issues and MANY fundamental mistakes ** I have in the past made pull requests into standard Arduino code base for long missed errors.

> Maybe I got your question wrong, but, the delay itself isn't for the timing of the pwm signal. We still use a 50hz signal, generated using XMC libraries. The delay is so the motor, as a electromechanical part which has some delay (between getting the signal, processing it, generating induction, etc), can get to the position before our mcu sends another one. It's an arbitrary value, maybe, but that's what the default example code uses, and I was not going to change it to another arbitrary value.

I understand very fully about servo response time the code is running too fast, the update rate is **NOT** an aritrary value unless you want **MISSED** steps.

Example of missed steps
- 5 iterations of the loop to sweep (or pot) is 5 * 15 = 75ms
- 75 ms with PWM/servo PHYSICAL update rate of 20 ms is 3.75 ACTUAL updates to servo so at least one if not two steps are MISSED

> Now, those are examples, the library itself doesn't have any delay of this kind. In those examples we want:
> -Sweep: The motor should move back and forth in a smooth stepping. Note that an example should always work for the user, as he may not be that experienced. So, maybe his servo isn't fast enough to change 1 degree in less than 15ms, that's why we use a delay here.
> -Knob: Here we want the servo to move with the value it read from a potentiometer, again, we use a delay so we don't spam the servo with like 200 changes in the PPM before it even got to the end of the 20ms wave (which may happen if the user starts to rotate the knob fast).

I know how the examples work badly, the electronics for servos (used in commercial projects on many micros), the different ranges for servos and some electronic devices like high current quadcopter brushless motor controllers that operate on servo signals.
 
> But again, the library itself is fully based on hardware pwm generation, and doesn't have a single delay.

I am well aware of the code and the inner depths of how the AVR and simialr versions work (badly)
 
> Now, about the "this->" stuff, this is my preference when writting any object oriented language (java, c++, etc). I use it wherever possible as it makes easier for me to read my own code. I change it as I see this isn't how code is done here, no problem. It isn't about "copying errors unknown".

Please check my readme's there are many errors in the AVR code that are not covered in their reference
 
> About the 544 to 2400 default values, you are right. This was what I used for my own project, I forgot that I changed that. I'll change it back.

I do not see where you setup the servo FREQUENCY in the code, and some places have maximum pulse of 20000 which as microseconds is the servo frequency as time

techpaul:
> Actually, about the default ms values, idk what Arduino does, but the industry default values are between 1000 and 2000ms for a single channel PPM servo.

NO that is what a lot of people use to limit on the SERVO it is **NOT the limit of the interface**

> There are many sources for this, you can check in the web, but here is one with it very detailed: https://dronebotworkshop.com/servo-motors-with-arduino/
> 
> Also from wikipedia: https://en.wikipedia.org/wiki/Servo_control#/media/File:Sinais_controle_servomotor.JPG

Wrong way round the standard for radio control and the maximum width is to do with transmitters and receivers and the PPM of the individual servo channles. this has always been an absolute maximum of 0.5 ms to 2.5ms plus sync pulse in 20 ms, which ALWAYS made 8 channel radio conytrol systems difficult.

Many SERVOS that RECEIVE their servo channel pulse once every 20 ms at fastest have these ays limits of 1ms to 2 ms for ease of the hardware/software/and channel compression for using methods such as PCM and other radio links. I have dealt with many transmitters, receivers, servos, or servo interface devices that have ranges 0.6ms to 2.4 ms. That is why many radio control transmitters(controllrs) had adjustable end stops on their joysticks to match servos.

The adding of min and max to the attach method was to limit the signal timing to match the servo limits for an individual servo, the interface has a spec of 0.5 to 2.5ms ABSOLUTE maximum.

Attach having min_angle and max_angle is different to standard Arduino and can cause conflicts in timing to degree and vice versa translations, in some cases

danielstuart14:
Here is the frequency line, set to 50hz: https://github.com/Infineon/XMC-for-Arduino/blob/8c6693e1042c7f796153a6e83651e1345938c8c1/libraries/Servo/src/Servo.cpp#L34

About the maximum and minimum, indeed the interface supports a broader timing, but, at least as what I think, the default values should work with what most people use, which are servos like the ones I described.

techpaul:
There are some other issues and I think at least one mapping range looks wrong, I will see about getting time tomorrow to sort out a hardware test as some things need to be checked on an oscilloscope.

techpaul:
As far as I can see this version suffers the same errors as standard Servo of 
1- able to create multiple classes on same pin
2- no error or invalid setup for assigning a servo to a NON-PWM pin (see return values for setAnalogWriteFrequency and analogWrite)

techpaul:
Having been running tests using this library AND one I was creating see (shortly when upload complete) https://github.com/techpaul/XMC_Servo Under examples are a couple of test programmes for checking fixed positions are stable and accurate one by servo calls other by direct analogWrite functions

There is an issue that affects **ALL** versions comanded timings are not achieved by an amount that will cause mid-point values (1500) to be short enough to cause "creep" on many electronic servo interface devices. I have checked through maths and calculations ALL the way down to register writes and issue is not library, core but trimming of hardware and calibrating clocks,

Following is table of results measured on scope using tests above on a few boards these from an XMC1300 Boot Kit

Position measured | Servotest | % error | PWMTest| % error
:--|--:|:-:|--:|:-:|
Refresh Frequency |  50.3573 | 0.715% | 50.3679 | 0.736%
Min - 544 | 540.2 | -0.699% | 539.5 | -0.827 %
Mid - 1500 | 1490 | -0.667% | 1489 | -0.733 %
Max - 2400 | 2383 | -0.708% | -- | -- |

Screenshot (with measurements) for Servo Test using 3 PWMs at same time
ServoTest

ScreenShot for PWMTest with two PWM outputs
PWMTest

Manually on PWMtest adjusting the duty cycle can get correct timing but amount is a complex stage of calculation that depends on measuring any implementation and what other factors are in play. Changing duty cycles as follows

Position measured | Expected duty | Trimmed Duty
:--|--:|:--
Min - 544 | 1782 | 1907
Mid - 1500 | 4915 | 5259

Using trimmed values gives correct results as shown in image
PWMTest2

From the errors and percentage factors it is clear that the internal resonator (**NOT** crystal) is OFF frequency. Back calculating having double checked actual results based on 32 MHz board with PCLK being supposedly 64 MHz

Prescaler is 32
Period count value for refresh frequency is 40000 

Back calculation from measered refresh frequency gives the main clock (MCLK and PCLK) are off and gives PCLK varying around  64,457,300 and 64,470,950 Hz instead of 64,000,000 Hz.

This means the PCLK is fluctuating all thye time suggesting variation between batches of chips or even different devices. There is possibility within the devices to do fractional adjustment of main oscillator and also trimming due to die temperature changes. However I am not sure what will happen between devices and batches for adjuting main oscillator, also how often to temperature calibrate (and how) at regular intervals. Having dealt with wide temperature range circuits getting clocks right over a temperature range is quite tricky and needs batch testing

If you are able to get actual measurements of timing for fixed values and refresh frequency with either library this would help to iron out how to get this correct.

techpaul:
@danielstuart14 

I am looking at main oscillator tuning and temperature adjust. First thing is change some standard software to allow negative temperatures to be returned and add as background task temp adjustment to main clock and test. To this end my intention is to do a NON-laboratory standard test, which will be good enough to prove if this works, basically run a test without software temp adjustment then run again with temp adjustment, on a selection of boards.

However this sort of test would be best performed on a variety and similar boards from hopefully different batches. I will be testing on
- XMC4700 Relax
- XMC1100 Boot Kit
- XMC1100 2GO
- XMC1300 Boot Kit

Test will be setup a 10KHz  50% duty cycle PWM and loop reading out die temperature.

Measure results using 
- scope/frequency counter to monitor output frequency at variety of temperatures
- Digital thermometer to measure PCB temp
- Software to spit out regular Die temp

Intention is to run test from Room Temp to -20C, which can be achieved by wires from PCB and measuring board on a surface, then placing board in a domestic freezer (these are capable of -20C to -24C. Noting readings at PCB temp of
20C to -20C as minimum range on all tests with 5C steps and note results on return to room temp. From this I hope to find if the PCLK can be made a lot tighter for more accurate Servo timings.

Can you also replicate this test?

If so -
- Do you have the equipment to do this?
- What boards do you have?

If this is not sufficient then I may have to look at dropping PCLK to 32MHz and a way to do fracional div calculations, for individual boards to be calibrated and a method to pass this into sketches

I have had to do this sort of system testing on devices monitoring all sorts of temp ranges but luckily not on the -200C to 600C tests.

danielstuart14:
@techpaul currently, because of the pandemia, I'm not able to test with the xmc 2 go I have at my university. But I have a 4700 relax here, which I can test it.
I have the necessary instruments, I'll try to reproduce it.

danielstuart14:
I'll close this pull request, we can talk privately.

PR #158, opened 2020-07-12
Die temp read negative temp and tidies

techpaul:
Change DeviceControl to allow reporting of negative temperatures, examine functions for later move to core for reading temperature.

Tidy some DeviceControl for simpler compiling and maintenance

Add etra tests for Servo/PWM/PCLK testing to see if better accuracy can be achieved


techpaul:
Most of this PR is needed for PCLK accuracy testing so will be merged now.

Some improvements will be beneficial to a few

PR #160, opened 2020-08-04
improved parameter handling 

mkiessling-ifx:
* Fixed setting parameter
* Added example to change parameters
* improved parameter handling

Arduino Interface:
* Provide option to set output device
* added function to dump parameters to "Print" Device

Arduino itself:
* added example on how to set parameters

No change to metadata of Lib!


techpaul:
Your commit coment
~~~
Note: after retesting this was needed, so not sure, why it worked with my tests before, since this was really missed.
further I need to include some XMCLib source files into my lib, since that is not pulled into the build, so I assume there was some change. Why that worked with my previous builds some days ago is also not 100% clear to me.
Resolved some warnings with the result callbacks
~~~
You should always test after a FRESH start of Arduino IDE, or change boards several times as many aspects of compile part results are cached during sessions to be sure always do test compile after restarting Arduino IDE.

The empty flags statement is needed so the IDE can ADD library specifics if architecture matches

You should be able to include the necessary XMCLib headers files by reference s long as you are using version 2.1.6, if the XMCLib source you are copying  is newer than 2.1.6, that would be a major upgrade and test cycle.

To see versions of board etc run tests/testdefines/testdefines.ino set for your board.

I would suggest you do an empty sketch test compile for your board variant as in 

1.  In Arduino IDE create new sketch
2. Ensure correct board set
3. Set preferences to Compiler Warnings ALL and verbose output on compilation
4. Do Verify/Compile and check for NO warnings at all

mkiessling-ifx:
Well, I added the flag statement, lib it is used in the mbed examples. 
So far, no other pre-compiled lib is used in this package. 
From Version 1.4 to Version 1.5 the variable "{compiler.libraries.ldflags}" was removed in the "recipe.c.combine.pattern" of platform.txt; So that after the update, the 0example for my lib does not build anymore. 
(I'm just discovering it by diffing the two versions separately.)

I just need to re-introduce it, but I just now noticed that the update was critical for my testing. 

I hope that works out now. Thank you for reviewing it!

techpaul:
I am reviewing from reading the code and build files, no other hardware etc in use

jaenrig-ifx:
So far the only change related to the platform.txt the deletion/addition ```{compiler.libraries.ldflags}```.  Is there @techpaul any reason to have removed that flag from 1.4.0 to 1.5.0 that you are aware of?

techpaul:
Not from memory as who did that or why.

PR #171, opened 2021-01-25
Fixes and raw data interface for BGT24LTR11-Pulsed-Doppler-Radar lib

mkiessling-ifx:
* rename board to "Radar BaseBoard XMC4700"
* fixes build and linker issues with usage of this library 
* added raw data interface
* updated examples
* added example to plot I/Q signal with build in Serial plotter


techpaul:
A couple of suggestions

pins_arduino.h missing definition
~~~
#define LED_BUILTIN xx  // Standard Arduino LED 
~~~
When running other code or examples some use this definition, use the same pin as one of your other LEDs.

In your examples and other files you use digital.write to set a specific level for LED on/off definitions in pins_arduino.h for LED being HIGH or LOW for a specific board mean you can use the LED library to make your examples more portable to other boards using same chipset. Only having to say 
~~~
Led.Red( ON );
or
Led.Red( OFF );
~~~
Is more portable across any Radar board, it takes care of does it need to be HIGH or LOW to turn on or off.


See LED library

mkiessling-ifx:
@techpaul: Thank you for the hint with Led-Object. That makes the examples really more expressive.

techpaul (approved):

mhollfelder:
Created new branch `24GHz-radar` for 24 Ghz Radar related changes and fixes before merging into develop.

PR #174, opened 2021-06-21
Fix typo specifying wrong input source for XMC1100 BootKit USIC

mtnmbr:
- when compiling with -DSERIAL_ONBOARD the define SERIAL_DEBUG becomes undefined.
   This leads to the use of P1_2 as inputsource of USIC0_CH1. The corresponding define
   specifies USIC0_C0_DX0_P1_2 which does not exist for XMC1100. Therefore
   the define has been changed to the correct CH1 define USIC0_C1_DX0_P1_2.

### DIFF
--- a/variants/XMC1100/config/XMC1100_Boot_Kit/pins_arduino.h
+++ b/variants/XMC1100/config/XMC1100_Boot_Kit/pins_arduino.h
@@ -211,7 +211,7 @@ XMC_UART_t XMC_UART_0 =
 #ifdef SERIAL_DEBUG
   .input_source_dx0     = (XMC_USIC_INPUT_t)USIC0_C1_DX0_P1_3,
 #else
-  .input_source_dx0     = (XMC_USIC_INPUT_t)USIC0_C0_DX0_P1_2,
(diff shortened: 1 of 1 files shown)


techpaul:
Sorry for delay. Can see the problem and will merge shortly

@mtnmbr @jaenrig-ifx Please note the problem arises from a PCB mistake with the software bodge to get around problem that someone designed the boards and connected two micros (the debugger and target) so their RX and TX pins were connected RX to RX and TX to TX. Hopefully someday we can get an updated PCB model that does not require this bodge

mtnmbr:
No problem! Thanks for merging the fix. What's the plan for the next release version? So the changes get pushed to the people working with the fixed releases on PlatformIO or in the Arduino IDE?

techpaul:
Unfortunately I am currently a contributor, it really is up to Infineon team to organise a release I make some changes and would need to review the release details to make sure it works correctly.

I have a few other things I need to find time to sort for other issues and try to get a board or two supported  Like XMC1400 Arduino Kit.

I will see if I can find some time to review release procedures and scripts. No promises when.

PR #190, opened 2022-06-29
Enable XMC1400 Kit for Arduino board into the Arduino IDE

boramonideep:
By creating this pull request you agree to the terms in CONTRIBUTING.md.
https://github.com/Infineon/.github/blob/master/CONTRIBUTING.md
--- DO NOT DELETE ANYTHING ABOVE THIS LINE ---

CONTRIBUTING.md also tells you what to expect in the PR process.

Description
This adds the XMC 1400 Kit for Arduino board into the XMC for Arduino BSP

Related Issue
#189 on github

Context
Please refer to internal ticket https://jirard.intra.infineon.com/browse/DESMAKERS-2490

PR #196, opened 2022-07-29
Tidy up variants and documentation

techpaul:
Removes unused and obsolete boards never released and tested

Improves documentation for LED library to support newer board support

Corrects polarity of LEDs for XMC4 series boards


jaenrig-ifx:
@techpaul

Any board or feature removal we are only considering for develop-2.x. 
Please remove such deletions, and we will evaluate the LED and documentation improvements.

Thanks a lot

jaenrig-ifx:
Hi @techpaul 

Sorry for the delay. 
I can see that the boards XMC1100 boot kit and the XMC1400 are changing the pinout from reset pin above. Those are not part of the Arduino UNO form factor and it is likely that not many users are already using them. But if that would be the case, it would break the compatibility with existing user application projects.
I suggest that such removal of reset pin from the available io pin numbers (including updated pinout diagrams) is pushed against the develop-2.x branch, and we add the improvements around hard_reset, LED and formatting are merged against the develop (1.x).

Thanks a lot!

techpaul:
@jaenrig-ifx 
Yes no and maybe is the short answer to that.

XMC1400 only last month released to develop,  has NOT been released to master and into the wild so which projects will be affected? Next to none, most of the pins are AD_AUX_x also referenced by A6 to A9 (the more likely toute), LED3 is most likely to be refenced by LED3, the only possible issue is int1 and as that is already not on a standard connector is already a compatability issue. On a not relased beyond develop recently.

XMC1400 potential issues NEXT to ZERO in the wild

XMC1100 MAYBE, but unlikely as most of the pins are AD_AUX_x (A6 and A7), three LEDs with LEDn definitions. (LED 1, LED2, LED6)

I personally have several little commercial (ONE off) test equipment in the wild using XMC1100 and a software update would have zero effect as those use A6, A7 , LEDn definitions

The likelihood of issues for XMC1100 is very small.

Anyway once the Reset pin has been assigned any operations to hadle the pin as GPIO will have zero effect as the Alternative Function of ERU input for Reset has priority in the hardware. Personally fixed function pin should not be in the pins available.

Leaving to when develop-2.x gets realeased in the wild INCREASES likelihood of issues later especially on XMC1400

techpaul:
XMC1400 merged 4 - JUL -2022 #190 

So been available in develop and possibly develop-2.x for very short time

jaenrig-ifx:
@techpaul true, the xmc1400 has not been yet merged into the master.
Okay, then I would say we take those changes for the xmc1400 arduino kit into develop(1.x), and just for the "maybes" (I also think there might be no cases) we can bring those changes of xmc1100 boot kit pinout ONLY to develop-2.x.

I know this is a bit of a pain to split those into several pull requests, but we are sure not the 1.x compatibility is not (at least consciously) broken. 

Thanks!

techpaul:
Will juggle commits shortly
Looking  back should have done this at V1.1.0 (April 2018) when it was one board before being used as template for XMC1400 and maybe others.

Hopefully we can set a style anything that is a fixed alternative function is not in digital pins  array etc.

variants/XMC1100/config/XMC1100_Boot_Kit/pins_arduino.h
```diff
@@ -88,21 +92,23 @@ extern uint8_t SCK;
 #define A6   6
 #define A7   7
 
-#define AD_AUX_1    24  // AD_AUX
```
jaenrig-ifx, line 91: So we keep these pinout assignment as it was for the 1.x and then these changes are done for 2.x or?
For the xmc1400 as mentioned, at it did not make it to the master, we can consider those changes ;)

techpaul:
I will get to it shortly, unfortunately had a family health emergency talking up a lot of my time this week

jaenrig-ifx:
No rush, just checking if we were still on the same page. I wish all goes well and there is full recovery from the health issue :)

techpaul:
> No rush, just checking if we were still on the same page. I wish all goes well and there is full recovery from the health issue :)

Got some time to complete it.

Unfortunately I have been busy with elderly parents organising care for father who has many problems, Alzheimer's, Chronic Kidney and a spread of cancer, so in this case no recovery just a waiting game and a lot of admin.

jaenrig-ifx (approved):
Thanks!

djaumann (approved):

OlafFilies (approved):

PR #200, opened 2022-08-28
Remove Reset pin from Digital Pin mapping XMC1100 Boot Kit

techpaul:
XMC1100 Boot Kit

Remove Reset pin from digital I/O mapping as cannot be used.
New XMC1100 Boot Kit pinout image in resources folder

Created this way after comments in #196


jaenrig-ifx (approved):

djaumann (commented):
Looks good :) Just one thing from my side - I wouldn't have removed the comments (pin numbers) behind the LED definitions. I think that makes it easier for the user to quickly see which LED is on which pin. But just a minor thing.

But they can check it of course in the pin mapping array.

techpaul:
> Looks good :) Just one thing from my side - I wouldn't have removed the comments (pin numbers) behind the LED definitions. I think that makes it easier for the user to quickly see which LED is on which pin. But just a minor thing.
> 
> But they can check it of course in the pin mapping array.

A few things

Firstly  in Arduino land you rarely talk P0.5 you talk abstracted digital pin 'n' which is already in the define. The port definitions are for lower levels after pin mapping talking to XMCLib etc.

Secondly if ANOTHER compiler upgrade causes problems like previously where sbrk got lost, comments on a #define can cause compilation errors if compiler stops doing substitute till a comment, as the substitution is often everything to end of line. Then you end up with lots of errors that need fixing due to ccompiler change.

General rule on defines in ANY flaviour of C, C++ C#...... do NOT put comments on same line as define, it can have hidden effects especially if a module is being included in a C and also C++ module, and different rules may apply.

You may have GOT AWAY with the practice so far as many madules exist as C or C++

djaumann (approved):

PR #225, opened 2023-01-11
Support for XMC2GO 32K Flash

devweblab:
Hi! Thank you for for this project and the arduino support! xmc2go works great.
I would like to mention that these boards are available with 32K or 64K flash.
Unfortunately you cannot explicitly order the 64K version! (Supplier confirmed that)
With Chip 032B instead of 064B you get an error when uploading >32k.
I have now 10 pieces and it is the 32K version only :(
(Other people i know received the 64K version)

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -91,6 +91,50 @@ XMC1100_XMC2GO.menu.LIB.DSPNN=ARM DSP / ARM NN Framework
 XMC1100_XMC2GO.menu.LIB.DSPNN.library.selected=-DARM_LIB_CMSIS_DSP -DARM_LIB_CMSIS_NN
 
 ####################################################
+XMC1100_XMC2GO_32K.name=XMC1100 XMC2Go (32K Flash)
+XMC1100_XMC2GO_32K.upload.tool=xmcprog
+XMC1100_XMC2GO_32K.upload.speed=115200
+XMC1100_XMC2GO_32K.upload.resetmethod=ck
+XMC1100_XMC2GO_32K.upload.maximum_size=32768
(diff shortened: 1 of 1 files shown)


techpaul:
I am NOT a staff memeber of Infineon just a contributor

Is this getting axtual XMC2GO boards with only 32k Flash or your own design board??

In my OPINION  this fixes HALF of the issue.

For comparison please look at XMC1300 variants in variants/XMC1300, there are a few linker_script_xxk.ld files for different flash sizes, and the relevant file is referenced in the boards.txt file for the specific board.

The entries in the boards.txt file mainly refer to flash size for actually flashing the device and have little effect on the compiling and linking.

Linker scripts give maximum flash size so any > flash size is caught at compile time.

Take a copy of linker_script.ld in variants/XMC1100 folder and call it linker_script_32k.ld, make the necessary size change in there, and alter boards.txt to referemce the linker_script_32k.ld this way you complete the fix.

I cannot comment on whether this will be incorporated into the repo as that is a choice of Infineon

devweblab:
Yes, it is the default/original xmc2go. Same packaging and board but different flash size. The old datasheet was for the 64k version only and the newer datasheet is for 32k or 64k.
Thank you very much. True, flash length in linker should be 0x8000 instead of 0x10000. Or is there anything else to change? I am not an expert, sorry.
With my changes the message in Arduino is better in my opinion: "Compilation error: text section exceeds available space in board" (if sketch is >32k)  instead of an upload error. Yes, up to infineon if they fix/change that.
Took me some time to figure out that the chip/flash size is the problem and not the code in the sketch.

PR #231, opened 2023-03-09
Changes in boards.txt for V2.x

boramonideep:

Description
Certain mistakes in boards.txt

Context
- Information in boards.txt file is verified. 
- One mistake was found and corrected.  
- Certain linker scripts which were not needed anymore in V2.x were removed. 

Testing
No testing required since the change was only in the package number of the XMC1100 in XMC2Go is changed. This would only change the define macros for the particular device type. There are no differences in those except for the pins that do not occur in one package. Hence, there is no need of explicit testing.


jaenrig-ifx (approved):

boramonideep:
Thanks @jaenrig-ifx  👍

djaumann (approved):

PR #232, opened 2023-03-09
Changes to boards.txt for V1.7

boramonideep:

Description
Certain mistakes in boards.txt

Context

- Information in boards.txt file is verified.
- Mistakes found and corrected, similar to V2.x
- Serial communication via PC was not working before on HBRIDGE2Go due to a wrong compile-time define. Corrected

Testing

- No testing required for the changes similar to V2.x since the change was only in the package number of the XMC1100 in XMC2Go is changed. This would only change the define macros for the particular device type. There are no differences in those except for the pins that do not occur in one package. Hence, there is no need of explicit testing.
- For the change related to HBRIDGE2Go, serial communication can be checked with a serial client on PC, using an example sketch.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -67,7 +67,7 @@ XMC1100_XMC2GO.build.mcu=cortex-m0
 XMC1100_XMC2GO.build.f_cpu=32000000L
 XMC1100_XMC2GO.build.board=ARM_XMC
 XMC1100_XMC2GO.build.board.version=1100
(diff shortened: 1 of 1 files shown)


jaenrig-ifx (approved):

boramonideep:
Thanks @jaenrig-ifx

djaumann (approved):

PR #249, opened 2023-07-16
- fixed implementation of transfer(buf, count) to match standard Ardu…

RudolphRiedel:
…ino implementation

Description
This fixes the implementation of transfer(buf, count)
from
void transfer(uint8_t* buf, uint8_t count)
to
void transfer(void *buf, size_t count)

As the latter is what is used by the standard Arduino API, see
https://github.com/arduino/ArduinoCore-API/blob/master/api/HardwareSPI.h

On top it also removes the limitation in length to 256 bytes while not using an extra buffer anymore that is allocated from the stack.

### DIFF
--- a/libraries/SPI/src/SPI.h
+++ b/libraries/SPI/src/SPI.h
@@ -150,7 +150,7 @@ public:
 
     uint8_t transfer(uint8_t data);
     inline uint16_t transfer16(uint16_t data);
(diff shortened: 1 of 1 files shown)


jaenrig-ifx (approved):
Thanks @RudolphRiedel ! 
That was necessary.

boramonideep (approved):
Looks good

OlafFilies (approved):

PR #255, opened 2023-09-19
Sd card change

OlafFilies:
This changes will make the SD card on XMC4700 RelaxKits available. Therefore the right SPI pins must be set and a macro for the XMC4700 RelaxKit has to be set. 

Keep in mind that only certain versions of the XMC4700 have the on board SD card slot. Therefore switch on the macro in SPI.h if you don't need it. 

This change has no effect on any other XMC series.


ederjc (approved):

boramonideep (approved):
looks  good

jaenrig-ifx:
Looks good. Is is also required to add the menu option in boards.txt?

jaenrig-ifx (approved):

PR #261, opened 2024-01-15
Fix xmc flasher

boramonideep:

Description
Fixing certain issues with the new XMC Flasher script


OlafFilies (approved):
Looks good to me so far.
Question,  this works now on all Python 3 minor versions, including the latest?

boramonideep:
> Looks good to me so far. Question, this works now on all Python 3 minor versions, including the latest?

Tested locally with 3.9, 3.10, 3.11 and 3.12 (latest). For a thorough testing, we could use something like `tox` to test with more versions without installing all of them locally.

PR #271, opened 2024-04-03
XMC CAN Sending & Unity starting point 

LinjingZhang:

Description
Enable all CAN basic function s

Related Issue
n/a

Context
https://jirard.intra.infineon.com/browse/DESMAKERS-3775 (CAN API)
https://jirard.intra.infineon.com/browse/DESMAKERS-3821 (Unity Test) - Sorry again I will try to separate the PR next time... 🐱 

The XMC CAN module works a bit different: It will not receive all data frames and then filter them. Instead, it only accept data frame with defined ID/ ID Mask. Spent two days reading the user manual and experimenting, no success

⚠️ Therefore the extra setDeviceId() function is added and I could not implement the filtering feature.

### DIFF
--- a/libraries/CAN/src/CANXMC.cpp
+++ b/libraries/CAN/src/CANXMC.cpp
@@ -14,7 +14,7 @@ namespace ifx
         .can_mo_type = XMC_CAN_MO_TYPE_RECMSGOBJ,
       };
 
-   XMC_CAN_MO_t CAN_msg_tx = 
+  XMC_CAN_MO_t CAN_msg_tx = 
 {       
     .can_mo_ptr = (CAN_MO_TypeDef*)CAN_MO1,
     {0xFF,  XMC_CAN_FRAME_TYPE_STANDARD_11BITS, XMC_CAN_ARBITRATION_MODE_ORDER_BASED_PRIO_1}, // {can_identifier, can_id_mode, can_priority}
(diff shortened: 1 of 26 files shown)


LinjingZhang:
### Example program
make FQBN=Infineon:xmc:XMC4700_Relax_Kit PORT=COM20 CANReceiver flash monitor

make FQBN=Infineon:xmc:XMC4700_Relax_Kit PORT=COM20 CANReceiverCallback flash monitor

make FQBN=Infineon:xmc:XMC4700_Relax_Kit PORT=COM24 CANSender flash monitor

only example flash is enable, unit test in development

libraries/CAN/src/CANXMC.cpp
```diff
@@ -33,9 +42,14 @@ namespace ifx
   }
   CANXMC::~CANXMC() {}
 
+  int CANXMC::setDeviceId(long id = 0x12) { // TODO: delete in the future!
```
jaenrig-ifx, line 45: This id should not have a default value I guess?

jaenrig-ifx (changes requested):
Getting there! 👍 

Unity should not be part of the code sources... but specific configuration files for this project. The rest should be able to be pulled-in with some external dependency management system providing the required unity version.

libraries/CAN/src/CANXMC.cpp
LinjingZhang: fix

jaenrig-ifx (approved):
Pending changes postponed. Future PRs.

PR #282, opened 2024-06-23
Fix for .miso_config / update of package.json

RudolphRiedel:

Description
A couple of variants/.../config/.../pins_arduino.h was using this for .miso_config:
    .miso_config      = {
        .mode = XMC_GPIO_MODE_INPUT_TRISTATE,
        .input_hysteresis = XMC_GPIO_INPUT_HYSTERESIS_STANDARD
    },

And since that skipped .output_level of the struct XMC_GPIO_CONFIG, building for XMC1100_XMC2GO failed with:
"sorry, unimplemented: non-trivial designated initializers not supported."

As the same project built just fine for the XMC4700_Relax_Kit, I compared the files and found this difference.
Adding .output_level for an input pin seems to be wrong and XMC_GPIO_INPUT_HYSTERESIS_STANDARD seems to be the default anyways, so I tried removing that line and it built again.

I also updated package.json from "version": "2.1.0" to "version": "3.1.0" while also adding a couple more lines that are found in the package.json that is installed by PlatformIO.
Ok, 3.1.0 is not really correct, this should be changed to whatever the next release is supposed to be so that it is correct in the release package - but I can not know if the next release will be 3.1.1, 3.2.0 or 4.x or something else entirely.

While this works to use the latest version in PlatformIO:
platform_packages = framework-arduinoxmc @ https://github.com/Infineon/XMC-for-Arduino.git
When listing the dependancies it is listed as framework-arduinoxmc @ 2.1 - but it uses the latest files, so 3.1.0+ actually.

Related Issue
-

Context
I am using PlatformIO to build my project and it is using arm-none-eabi-g++.
I had to manually convince PlatformIO to use the latest version of XMC-for-Arduino, the platform "Infineon XMC" is still using "framework-arduinoxmc" in version 1.4.0 and this one uses a different pins_arduino.h which does not even have the SPI config.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,15 @@
 {
     "name": "framework-arduinoxmc",
+    "version": "3.1.0",
     "description": "Integration of Infineon's XMC microcontrollers with the Arduino framework",
     "url": "https://www.infineon.com/cms/en/product/microcontroller/32-bit-industrial-microcontroller-based-on-arm-cortex-m/",
-    "version": "2.1.0"
-}
+    "keywords": [
+        "framework",
+        "arduino",
+        "infineon"
+      ],
+      "repository": {
+        "type": "git",
+        "url": "https://github.com/Infineon/XMC-for-Arduino"
+      }
+    }
--- a/variants/XMC1100/config/XMC1100_Boot_Kit/pins_arduino.h
+++ b/variants/XMC1100/config/XMC1100_Boot_Kit/pins_arduino.h
@@ -96,7 +96,7 @@ extern uint8_t SCK;
 #define MASTER_SCLK     PORT0, 8
 #define MASTER_WACLK    PORT0, 9
 
-// Uncomment this to use the second I2S interface,
+// Uncomment this to use the second I2S interface, 
 // uses some of the same pins as SPI!
 // /*U0C0*/
 // /*DX0C(Input) -> P1.0*/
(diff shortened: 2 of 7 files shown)


package.json
LinjingZhang, line 1: Thank you for the update

variants/XMC1100/config/XMC1100_Boot_Kit/pins_arduino.h
LinjingZhang, line 1: I may not be able to merge this change
1. in platfomIO, we are using a lower version of the gcc toolchain, so it can't be compiled. if you try modifying the `package[toolchain-gccarmnoneeabi][version]` in `.platformio\platforms\infineonxmc\platform.json` to `1.100301.220327`, you should be able to compile the code without this change.
2. The code contains many formatting changes. Thanks for the reminder, we will automate the c/c++ formatter as soon as possible

variants/XMC1100/config/XMC1100_XMC2GO/pins_arduino.h
LinjingZhang, line 1: same for all pins_arduino.h

LinjingZhang (changes requested):
Thank you very much for the PR. very sorry for the late reply.
hope you can continue to contribute to this repo 😄

variants/XMC1100/config/XMC1100_Boot_Kit/pins_arduino.h
RudolphRiedel: 1) I see, so the correct fix would be to update the gcc toolchain that platformio is using
2) I saw that more than I intended got changed, but this was not formatting, only removing useless trailing spaces

So, yeah, if this can be resolved with an update to platformio-infineonxmc, just kill this pull-request. :-)

RudolphRiedel: I was not sure how, but luckily I did the changes in two commits and just reverted the second commit.

LinjingZhang (approved):

PR #289, opened 2024-07-11
Fix/release process

LinjingZhang:

Description
Fix the release script error

Related Issue
release workflow is not working 

Context
release workflow is not working , we have to revert the files in 3.1.0 release.
The problem is in release.py line 76 (old version), where the json arrays are incorrectly combined.

### DIFF
--- a/.github/scripts/release.py
+++ b/.github/scripts/release.py
@@ -1,7 +1,7 @@
 
 import argparse, copy, hashlib, json, re, requests, os, shutil
 
-version = '0.2.0'
+version = '0.2.1'
 
 xmc_ino_root_path = os.path.relpath(os.path.join(os.path.join(os.getcwd(), os.pardir), os.pardir))
 build_dir_name = 'pkg_build'
@@ -79,7 +79,7 @@ def set_new_platform_data_fields(platform_data_index, pkg_name, version, reposit
     platform_data['size'] = str(get_package_size(os.path.join(pkg_assets_build_path, archive_file_name)))
 
 def add_platform_to_package_index(pkg_index, platform):    
-    pkg_index['packages'][0]['platforms'].insert(1, platform)
+    pkg_index['packages'][0]['platforms'].extend(platform)
 
 def make_package_index_file(pkg_index):
     pkg_index_json_obj = json.dumps(pkg_index, indent=2)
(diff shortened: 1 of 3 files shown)


.github/scripts/release.py
```diff
@@ -60,40 +60,54 @@ def get_package_sha256(pkg):
...
+    set_new_platform_data_fields(local_package_index, pkg_name, version, repository)
+    # get old package array
+    old_platform_data = get_platform_data_struct_copy(latest_package_index)
+    # append to local package index
+    add_platform_to_package_index(local_package_index, old_platform_data)
+    make_package_index_file(local_package_index)
+
+def build_release_assets(version, repository):
+    if os.path.exists(pkg_assets_build_path):
+        os.system("rm -rf "+pkg_assets_build_path)
```
jaenrig-ifx, line 105: Why using rm to delete the package? 
This will make this script only unix compatible.

.github/scripts/release.py
```diff
@@ -1,7 +1,7 @@
 
 import argparse, copy, hashlib, json, re, requests, os, shutil
 
-version = '0.1.0'
+version = '0.2.1'
```
jaenrig-ifx, line 4: 0.2.0?

jaenrig-ifx (changes requested):
👍 !

.github/scripts/release.py
LinjingZhang: The old script with the error was 0.2.0, but since it has been reverted, it is now invisible. should I change to 0.2.0?

jaenrig-ifx: If there was not really any release with 0.2.0 (as it was rolled back), then 0.2.0 should be the next or?

LinjingZhang: fixed

.github/scripts/release.py
LinjingZhang: I was testing it in wsl because the runner is also unix...
I can also remove it. This is easy for local test, but is most likely not needed for workflow  ( the path is clean

jaenrig-ifx: You don´t need to remove it, but instead use python apis to remove filesystem folders and files. That way we keep it cross platform.

jaenrig-ifx (approved):

PR #291, opened 2024-07-23
Final check before XMC1400 BSP/ CAN Lib release

LinjingZhang:

Description
Details in https://jirard.intra.infineon.com/browse/DESMAKERS-3943
code cleaned , doc updated. example tested. 

Related Issue
https://jirard.intra.infineon.com/browse/DESMAKERS-3943

Context
I added the unity structure I had before.
@jaenrig-ifx  I changed the release script a bit, so the test files and make files will not be included in release. And make is available for CICD :D


.clang-format
```diff
@@ -0,0 +1,22 @@
+---
+# We'll use defaults from the LLVM style, but with 4 columns indentation.
```
jaenrig-ifx, line 2: How to run this?
While this is not built-in automated check in the project, at least, we should add some documentation on how to locally run this formatter or check the format.
The truth is that without ci/cd check, the formatting won´t be ensured for every contribution. And if that is not the case... this is yet not very useful.

libraries/CAN/Makefile.test
```diff
@@ -2,12 +2,23 @@
 
 
 ### Example program
+## XMC4700_Relax_Kit
```
jaenrig-ifx, line 5: How to run this? 
Where is the makefile? etc?

jaenrig-ifx (changes requested):
Maybe we can document briefly those aspects of the formatting and the tests?
The CAN functional changes I have no judgment criteria. I expect it has been verified in the best way possible (manually).

.clang-format
LinjingZhang: .clang_format removed!

libraries/CAN/Makefile.test
LinjingZhang: I will keep this for now

.github/scripts/release.py
```diff
@@ -37,8 +37,12 @@ def build_package(pkg_name):
         'README.md'
     ]
 
+    ignore_pattern = shutil.ignore_patterns()
+
     for dir in dirs_to_copy:
-        shutil.copytree(os.path.join(xmc_ino_root_path, dir), os.path.join(pkg_build_path, dir))
```
jaenrig-ifx, line 41: Increase the version of the script to 0.2.2?

jaenrig-ifx (approved):

.github/scripts/release.py
LinjingZhang: The current version in master is 0.2.0. So 0.2.1 should be correct 🥲 But I don't know why the git log is so strange. We revert this file some times a go....
But in final merge you will see the script is correctly versioned

jaenrig-ifx: if that is the case 👍

PR #301, opened 2024-09-13
RTD Built in Libraries & Arduino Variations

CaGriesser:

Description
New changes in the Read the Docs Documentation. Added Built in Libraries and Arduino Varations. (Arduino Variations are not checked or changed in any way, it ist the same like in the wiki documentation)


jaenrig-ifx (approved):
Previous resolving the conflict. It looks goods 👍 

In general I see a potential improvement when we migrate ArduinoCore-API, and then resolve some of the variations in favor of interoperability with the API. We could try minimize any variation which is not a "feature extension".
But that isn´t part of the docs.

Also ideally in future, the built-in libraries/example support per board could be generated dynamically from ci/cd build/test matrices.

PR #305, opened 2024-09-24
Fix: Cannot find JLink device in Linux 

LinjingZhang:

Description
Error when flash in Linux. Adapt Regex expression
Thanks Olaf for finding this bug and providing a workaround

Related Issue
JLink device name is different for different OS.

Context
Tested on all OS

- [x] Windows
- [x] Linux
- [x] MacOS


ederjc (approved):
Tested and works fine, thanks for the quick fix!

PR #313, opened 2024-11-25
Feature/unity setup reuse

LinjingZhang:

Description
Provide the information we need to review your PR. What problem does the pull request solve? "Bug fix" is not a good description.

Related Issue
If you opened an issue before creating the PR, point to it here.

Context
What do we need to know about your development environment, tools, target, and so on. Screenshots are always helpful if there is a UI element to this PR.


bargfred (approved):

cores/HardwareSerial.cpp
```diff
@@ -1,234 +1,239 @@
...
-            // Mask off transmit interrupt so we don't get it any more
-            XMC_UART_CH_DisableEvent(_XMC_UART_config->channel, XMC_UART_CH_EVENT_TRANSMIT_BUFFER);
-        }
-    }
-}
-
-//****************************************************************************
-//                                 END OF FILE
-//****************************************************************************
+/*
```
jaenrig-ifx, line 1: What has changed here?
Seems all deleted and added again. Not sure how the code is being handled git-wise. But we should try to make these diff show the actual changes, to allow us review effectively, and focus on the changes.
Also for the hil-unity-checks..yml file.

tests/Makefile.test
```diff
@@ -0,0 +1,23 @@
+### Build commands
+### ! Please note that you need to change the port and path of Unity as appropriate !
+
+### Unit tests
+
+##CAN
+
+# 1 board, no wire
+make FQBN=Infineon:xmc:XMC1400_XMC2GO PORT=COM42 UNITY_PATH=\Unity test_can_single monitor
```
jaenrig-ifx, line 9: If this is part of the arduino-core-test, should it be (documented) there?

tests/hil-unity-checklist.yaml
```diff
@@ -0,0 +1,35 @@
+Infineon.xmc.XMC4700_Relax_Kit:
```
jaenrig-ifx, line 1: This is used by the hi l.yml worfklow I guess? 
Shall this be in future here or in the arduino-core-tests repo? 
If it refers to CAN and Wire I am not sure if this is longer of this domain...
But okay, let´s not worry to much about this one. Things we need to think about.

jaenrig-ifx (changes requested):
Two more things:

* The submodule should not be called "unity" any longer. It is now currently named as "arduino-core-tests".
The _de facto_ convention for other arduino cores is to have such external libraries or utilities in a "_extras/_" folder.
Not sure if that is creating any trouble in case some xmc-for-arduino specific tests (under tests/) needs to be called or run from that folder (and rely on the "arduino-core-test" in another path).

* Are the I2C and CAN tests already running in the HIL? Is the HIL failing due to these changes? 
If it is not failing due to these changes fine, but you should be able to run locally the desired tests after all these files reorgs and changes. Is that the case?

cores/HardwareSerial.cpp
LinjingZhang: I need to disable autocrlf for other project. I think problem is EOL inconsistency.
need  .gitattributes/ inputs from Architecture board 😉 
image

tests/Makefile.test
LinjingZhang: It's just a make command for developer reference. Removed considering that this is strongly project related.
Can also move back

tests/hil-unity-checklist.yaml
LinjingZhang: @OlafFilies

cores/HardwareSerial.cpp
LinjingZhang: fixed and added .gitattributes.
This timeout added for fix one of the i2c test.

jaenrig-ifx (approved):

PR #314, opened 2024-11-28
Adapting timeout while-loop in Wire.cpp

Frederik9128:
Error: Wire.cpp got stocked in the while-loop for the timeout since the timeout-condition was not defined as it should be per definition. It was not possible to leave the while-loop after the timeout-variable reached 0. 

Change: Now, the checking condition was reduced to the essentials (it was wrong before) and another if-condition, which checks if the timeout-variable reached 0, was added inside the while-loop. This leads to a break.

Testing: The new wire.cpp was tested with the KIT_XMC_2GO_XMC1100_V1 & KIT_XMC47_RELAX_LITE_V1. Both worked.


ederjc:
Deprecated by https://github.com/Infineon/XMC-for-Arduino/pull/317

PR #322, opened 2025-01-20
Bugfix/desmakers 4270 serial.print is not allways working correctly in xmc for arduino version 3.5.0

LinjingZhang:

Description
Removed the transmit ringbuffer in order to quickly fix this Serial.print().
More changes will be made when replace coreAPI.

Related Issue
Please check DESMAKERS 4270

Context
For UART transmission, the original XMC4Arduino implementation was very confusing. Although our XMC USIC already comes with a hardware buffer and the xmclib API has a wait feature to block tx, we also use an additional software Ringbuffer. So We remove the ringbuffer and simply use the XMC_UART_CH_Transmit API (When FIFO is not configured, the API waits for the TBUF to be available. This makes the execution a blocking call).


ederjc (approved):
Please squash & merge :)

PR #324, opened 2025-02-10
CI: Update compile-platform-examples.yml

LinjingZhang:

Description
Compilation test is not running due to deprecation of upload-artifact.

Related Issue
https://github.com/actions/upload-artifact/issues/478
It seems that upload-artifacts v4 does not allow all outputs in the build matrix to be uploaded to one folder. 
So now artifact is uploaded under different board names.

Context

### DIFF
--- a/.github/workflows/compile-platform-examples.yml
+++ b/.github/workflows/compile-platform-examples.yml
@@ -194,9 +194,14 @@ jobs:
           enable-deltas-report: false
           sketches-report-path: ${{ env.SKETCHES_REPORTS_PATH }}
 
+      - name: Modify the environment variable
+        run: |
+          BOARD_NAME=$(echo ${{ matrix.board.fqbn }} | sed 's/:/-/g')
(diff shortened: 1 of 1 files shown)


.github/workflows/compile-platform-examples.yml
```diff
@@ -194,9 +194,14 @@ jobs:
           enable-deltas-report: false
           sketches-report-path: ${{ env.SKETCHES_REPORTS_PATH }}
 
+      - name: Modify the environment variable
```
jaenrig-ifx, line 197: Is this step required? Can you just not use ${{ matrix.board.fqbn }} already in the naming of line 206?

LinjingZhang: no unfortunately colon is not allowd in the name so I have to add this step.

jaenrig-ifx: All good, anyhow, this action is not currently used in the arduino-core-tests. And might not be used. 
Still, we are using this action for uploading things in other places. So good to know they have change this behaviour.

PR #325, opened 2025-02-10
Bugfix/desmakers 4316 spi normal pins 11 12 13 not functioning on xmc4700 relax kit

LinjingZhang:

Description
Solve the problem that the default spi is overwritten when the XMC4700 relax kit sets the SD card PIN to the default value.
Tested the default SPI pins and the SD card function, both work fine!

Related Issue
https://jirard.intra.infineon.com/browse/DESMAKERS-4316

Context

### DIFF
--- a/libraries/SPI/src/HW_SPI.cpp
+++ b/libraries/SPI/src/HW_SPI.cpp
@@ -29,10 +29,9 @@
 
     // swap SPI_default and SPI_for_xmc_SD if desired by user
     #if defined(USE_XMC_RELAX_KIT_SD) && defined(XMC_SPI_for_xmc_SD)
-SPIClass SPI(&XMC_SPI_for_xmc_SD);
-        #if (NUM_SPI > 1)
-SPIClass SPI1(&XMC_SPI_default);
-        #endif
+SPIClass SPI(&XMC_SPI_default);
+SPIClass SPI1(&XMC_SPI_for_xmc_SD);
+
 uint8_t SS = PIN_SPI_SS_SD;
 uint8_t MOSI = PIN_SPI_MOSI_SD;
 uint8_t MISO = PIN_SPI_MISO_SD;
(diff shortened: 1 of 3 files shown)


libraries/SPI/src/HW_SPI.cpp
```diff
@@ -29,24 +29,18 @@
...
-        #if (NUM_SPI > 1)
-SPIClass SPI1(&XMC_SPI_default);
-        #endif
-uint8_t SS = PIN_SPI_SS_SD;
-uint8_t MOSI = PIN_SPI_MOSI_SD;
-uint8_t MISO = PIN_SPI_MISO_SD;
-uint8_t SCK = PIN_SPI_SCK_SD;
+SPIClass SPI(&XMC_SPI_default);
+SPIClass SPI1(&XMC_SPI_for_xmc_SD);
     #else // normal behaviour
```
jaenrig-ifx, line 34: Is this formatted? with the clang?
The indentation of the preprocessor and not the rest of the code it is quite unclear. The code should be indented together with the preprocessor sections. If not better not to indent the nested pre-processor directives.

.github/workflows/compile-platform-examples.yml
```diff
@@ -194,9 +194,14 @@ jobs:
           enable-deltas-report: false
           sketches-report-path: ${{ env.SKETCHES_REPORTS_PATH }}
 
+      - name: Modify the environment variable
```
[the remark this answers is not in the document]
LinjingZhang: otherwise compilation tests are not working. will first merge the other one to master

libraries/SPI/src/HW_SPI.cpp
LinjingZhang: Yes, this is actually after format, otherwise clang-format will report an error.
I can look into clang config. i also think this part is not clear enough

jaenrig-ifx: Yep, we have to configure that properly... But okay, that is not in the scope of this PR, so nothing to be changed for this PR.

jaenrig-ifx (approved):

PR #331, opened 2025-03-18
Desmakers 4279 rename the metafile vendor from infineon to infineon

LinjingZhang:

Description
rename the metafile vendor from infineon to infineon

Related Issue
Desmakers 4279

Context
Also need to adapt the documentation.
Will update the other PR.

### DIFF
--- a/.github/workflows/compile-platform-examples.yml
+++ b/.github/workflows/compile-platform-examples.yml
@@ -35,7 +35,7 @@ jobs:
 
       matrix:
         board:
-          - fqbn: Infineon:xmc:XMC1100_Boot_Kit
+          - fqbn: infineon:xmc:XMC1100_Boot_Kit
             i2s: true
             dieTemp: true
             heapMem: true
@@ -45,7 +45,7 @@ jobs:
             dma: false
             alarmRtc: false
             can: false
-          - fqbn: Infineon:xmc:XMC1100_XMC2GO
+          - fqbn: infineon:xmc:XMC1100_XMC2GO
             i2s: true
             dieTemp: true
             heapMem: true
@@ -55,7 +55,7 @@ jobs:
             dma: false
             alarmRtc: false
             can: false
-          - fqbn: Infineon:xmc:XMC1300_Boot_Kit
+          - fqbn: infineon:xmc:XMC1300_Boot_Kit
             i2s: false
             dieTemp: true
             heapMem: true
@@ -66,7 +66,7 @@ jobs:
             dma: false
             alarmRtc: false
             can: false
-          - fqbn: Infineon:xmc:XMC4200_Platform2GO
+          - fqbn: infineon:xmc:XMC4200_Platform2GO
             i2s: false
             dieTemp: false
             heapMem: false
@@ -77,7 +77,7 @@ jobs:
             dma: false
             alarmRtc: true
             can: true
-          - fqbn: Infineon:xmc:XMC4400_Platform2GO
+          - fqbn: infineon:xmc:XMC4400_Platform2GO
             i2s: false
             dieTemp: false
             heapMem: false
@@ -87,7 +87,7 @@ jobs:
             dma: false
             alarmRtc: false
             can: true
-          - fqbn: Infineon:xmc:XMC4700_Relax_Kit
+          - fqbn: infineon:xmc:XMC4700_Relax_Kit
             i2s: true
             dieTemp: true
             heapMem: true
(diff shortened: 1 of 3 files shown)


jaenrig-ifx:
Mmm, we are adding this as a 3.x change? 
I think this can break existing installation with the low case "I". Have this been tested? 
Any issue with bringing this change in 4.x? 

Also, as this is only for xmc .... ideally we could name the package_infineon_index.json -> package_xmc_index.json. But that will definitively break everything. But this will require the index url to be updated, otherwise the later versions will not be automatically available to the users. 
And for that we should add some notification in the build or such... So that they are aware and can upgrade to the new version.

ederjc:
Yes, this might have broader implications than we initially thought. Also regarding all build checks involving this core.
Is there a reason to merge this to 3.x except the compatibility problems with the PSOC core?

Also, please add required changes to `/tests/Makefile.test` in this PR, to make it complete.

LinjingZhang:
> Mmm, we are adding this as a 3.x change? I think this can break existing installation with the low case "I". Have this been tested? Any issue with bringing this change in 4.x?
> 
> Also, as this is only for xmc .... ideally we could name the package_infineon_index.json -> package_xmc_index.json. But that will definitively would be a "major" compatibility break. But this will require the index url to be updated, otherwise the later versions will not be automatically available to the users. And for that we should add some notification in the build or such... So that they are aware and can upgrade to the new version.

now merge to 4.0.0. Another reason to merge to 3.x first is, the compilation tests fails anyways in 4.0.0. I tested with 3.x so the changes should be fine be merged.
Other naming problem you mentioned are under : https://jirard.intra.infineon.com/browse/DESMAKERS-4159 ( you already created 😄 )

jaenrig-ifx:
> > Mmm, we are adding this as a 3.x change? I think this can break existing installation with the low case "I". Have this been tested? Any issue with bringing this change in 4.x?
> > Also, as this is only for xmc .... ideally we could name the package_infineon_index.json -> package_xmc_index.json. But that will definitively would be a "major" compatibility break. But this will require the index url to be updated, otherwise the later versions will not be automatically available to the users. And for that we should add some notification in the build or such... So that they are aware and can upgrade to the new version.
> 
> now merge to 4.0.0. Another reason to merge to 3.x first is, the compilation tests fails anyways in 4.0.0. I tested with 3.x so the changes should be fine be merged. Other naming problem you mentioned are under : https://jirard.intra.infineon.com/browse/DESMAKERS-4159 ( you already created 😄 )

You can also disable/comment for the time being all the tests which aren´t yet ready. And you can focus on the error of the modules/lib that you are working on.

ederjc (changes requested):
Sorry I missed this:
Please rework the commit messages to fit the format and correct spelling.
Our commit checks also seem to be not running properly for this PR?
https://github.com/Infineon/XMC-for-Arduino/actions/runs/13989397821

jaenrig-ifx:
what about changing the boards.txt names also snake_case?

LinjingZhang:
> what about changing the boards.txt names also snake_case?

https://jirard.intra.infineon.com/browse/DESMAKERS-4280
https://jirard.intra.infineon.com/browse/DESMAKERS-4282

jaenrig-ifx:
> > what about changing the boards.txt names also snake_case?
> 
> https://jirard.intra.infineon.com/browse/DESMAKERS-4280 https://jirard.intra.infineon.com/browse/DESMAKERS-428

It makes sense to bring all these changes at one in one PR, with all the compatibility breaking changes.

tools/verifygitlog.py
```diff
@@ -68,19 +68,11 @@ def diagnose_subject_line(subject_line, subject_line_format, err):
 def verify(sha, err):
     verbose("verify", sha)
     err.prefix = "commit " + sha + ": "
-
```
jaenrig-ifx, line 71: Can you explain the changes on this file? 
At some point, this will be moved to the reusable workflow for all cores. So it is good to align on the changes of these checks scripts.

LinjingZhang:
> > > what about changing the boards.txt names also snake_case?
> > 
> > 
> > https://jirard.intra.infineon.com/browse/DESMAKERS-4280 https://jirard.intra.infineon.com/browse/DESMAKERS-428
> 
> It makes sense to bring all these changes at one in one PR, with all the compatibility breaking changes.

OK I understand. Then we should align better during scrum planning next time.

tools/verifygitlog.py
jaenrig-ifx: okay, so this disables commit signing?

LinjingZhang: Disabled only for commit messages starting with ignore_prefixes = [“squash!”, “fixup!”, “amend!”, “WIP"].
Otherwise it will still be checked.

jaenrig-ifx: Okay, we will harmonize this when we make it reusable.

jaenrig-ifx (approved):
👍

LinjingZhang:
move the commit msg check to: https://github.com/Infineon/XMC-for-Arduino/pull/332

PR #335, opened 2025-04-04
Updated release to reusable workflow

jaenrig-ifx:

* Updated `release.yml `with reusable workflow from arduino-devops.


LinjingZhang:
nice!
Just one question, why did only the commit msg check run for this PR?

package/config.yml
```diff
@@ -0,0 +1,16 @@
+package-name: xmc-for-arduino
+include:
+  - cores
+  - libraries
+  - tools
+  - variants
+  - boards.txt
+  - platform.txt
```
LinjingZhang, line 8: To be on the safe side, I might consider referencing the previous script, and includes:
keywords.txt - Highlighting Keywords in the IDE
package.json - This is used in platform.io. Not sure if it continues to maintain but keeping it won't hurt

jaenrig-ifx: Right. I will add the platform.txt. 
I can also add the package.json. But is that working? We are not actively maintaining the integration, not sure if it still work? 
@ederjc maybe knows more?

jaenrig-ifx:
> nice! Just one question, why did only the commit msg check run for this PR?

I think because the worfklows are filtering only PR against master?

jaenrig-ifx:
> > nice! Just one question, why did only the commit msg check run for this PR?
> 
> I think because the worfklows are filtering only PR against master?

And if changes are made in certain paths and files.

package/config.yml
jaenrig-ifx: done!

LinjingZhang (approved):

PR #351, opened 2025-06-10
Core api update tone

dineshgit411:

Description
Tone function implemented using PWM method. In XMC we have two modes CCU4 and CCU8 . configured for all the PWM pins.

Related Issue
Removed delay in test_main.ino because test case not running in continuous loop. 
Context
Tested with 4700 Board, connected jumber wire in between interrupt pin and pwm pin.
image

### DIFF
--- a/cores/xmc/Arduino.h
+++ b/cores/xmc/Arduino.h
@@ -73,6 +73,8 @@ extern "C" {
 
 #define interrupts() __enable_irq()
 #define noInterrupts() __disable_irq()
+#define ENABLED 1
+#define DISABLED 0
 
 #define DEFAULT XMC_VADC_CHANNEL_REF_INTREF /*< Default ADC reference voltage type */
 
--- a/cores/xmc/Tone.cpp
+++ b/cores/xmc/Tone.cpp
@@ -1,5 +1,5 @@
 #include "Arduino.h"
-#define CCU8V2 // for CCU8 PWM pins configure
+
 XMC_PWM4_t *pwm4;
 
 class Tone {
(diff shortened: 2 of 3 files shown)


cores/xmc/Tone.cpp
```diff
@@ -0,0 +1,142 @@
...
+    void play(pin_size_t pin, unsigned int frequency, unsigned long duration) {
+        int pin_index;
+        if ((pin_index = scanMapTable(mapping_pin_PWM4, pin)) >= 0) {
+            XMC_PWM4_t *pwm4 = &mapping_pwm4[pin_index];
+            configureTone(pin, frequency);
+            XMC_GPIO_SetMode(pwm4->port_pin.port, pwm4->port_pin.pin,
+                             XMC_GPIO_MODE_OUTPUT_PUSH_PULL | pwm4->port_mode);
+            XMC_CCU4_SLICE_StartTimer(pwm4->slice);
+            // count pulses for stop the timer once reached required pulses
+            if (duration > 0) {
```
jaenrig-ifx, line 21: So if a duration is set the function is blocking or?

cores/xmc/Arduino.h
```diff
@@ -73,6 +73,8 @@ extern "C" {
 
 #define interrupts() __enable_irq()
 #define noInterrupts() __disable_irq()
+#define ENABLED 1
```
jaenrig-ifx, line 76: Where are these defines used?

tests/test_config.h
```diff
@@ -13,7 +13,7 @@
 #include <stdint.h>
 
 // Test Pin Definitions
-#define TEST_PIN_DIGITAL_IO_OUTPUT 2 // IO_4
-#define TEST_PIN_DIGITAL_IO_INPUT 3  // IO_3
+#define TEST_PIN_DIGITAL_IO_OUTPUT 3 // IO_4
```
jaenrig-ifx, line 16: Be careful which swapping pins if other tests rely on these pin values...

jaenrig-ifx (commented):
Looks very good 👍

tests/test_config.h
LinjingZhang: image
We checked this pin together and previous output pin could not do pwm therefore we modified it. For pwm and tone tests.
Involved tests such as digitalIO, interrupt will not be affected, since both can configure interrupt.
https://github.com/Infineon/arduino-core-tests/blob/main/src/corelibs/analogio/test_analogio_pwm.cpp#L22

LinjingZhang (changes requested):
I think the two issues juan mentioned need to be fixed
1. tone(duration) should not be blocking call -> maybe this is reason of scheduler code in timer module?  
2. true and false instad of ENABLED and DISABLED macro.

cores/xmc/Tone.cpp
dineshgit411: yep, I will change the logic. before, I used timer  but one test case got failed due to time variation(100ms)
I will revisit and change.

PR #357, opened 2025-06-30
Core api update iic

dineshgit411:

Description
Used Existing wire implementation and removed unnecessary lines and modified few things.
Consider 2 boards one is master and one is slave tested with test case and everything passed.

Related Issue
Need to test with Realtime I2C hardware.
Context
1.connected two 4700 board. consider one is master and one is slave
2.Tested with single board (two I2c specs)
image
image


jaenrig-ifx:
The Wire API needs to implement the following interface  -> https://github.com/arduino/ArduinoCore-API/blob/master/api/HardwareI2C.h

See psoc6 implementation as reference (inheriting from HardwareI2C class) --> https://github.com/Infineon/arduino-core-psoc6/blob/main/libraries/Wire/src/Wire.h#L11

Any cleanup or refactor in the code is also recommended. For example, like instead of implementing our own circular buffer, use the available in the core API https://github.com/arduino/ArduinoCore-API/blob/master/api/RingBuffer.h

dineshgit411:
Based on discussion , for this PR (read, write, available, beginTransmission , endTransmission, onReceive) functions are tested.  rest of the function need to implement in next sprint with another PR.

libraries/Wire/src/Wire.cpp
```diff
@@ -59,192 +14,133 @@ TwoWire::TwoWire(XMC_I2C_t *conf) {
...
 
-// Public Methods //////////////////////////////////////////////////////////////
-
 void TwoWire::begin(void) {
-    // Check if desire USIC channel is already in use
-    if ((XMC_I2C_config->channel->CCR & USIC_CH_CCR_MODE_Msk) == XMC_USIC_CH_OPERATING_MODE_SPI) {
-        SPI.end();
-    }
-
+    // To-Do for future work need to check
```
jaenrig-ifx, line 23: Feels like some workaround for some external application which was not calling SPI.end() before using the same pins as SPI. 
If we don´t have the concreate reasons why this was done, this can be removed. 
The user is responsible of calling SPI.end() before using I2C on the same pins.

libraries/Wire/src/Wire.cpp
```diff
@@ -476,34 +306,28 @@ size_t TwoWire::write(uint8_t data) {
...
-// must be called in:
-// slave tx event callback
-// or after beginTransmission(address)
 size_t TwoWire::write(const uint8_t *data, size_t quantity) {
     if (transmitting) {
-        // in master transmitter mode
         for (size_t i = 0; i < quantity; ++i) {
-            write(data[i]);
+            if(!write(data[i])) {
+                return 1;
```
jaenrig-ifx, line 313: "return 1" or "return i"? It should return the number of bytes written?

libraries/Wire/src/Wire.h
```diff
@@ -1,111 +1,57 @@
...
-// @Project Includes
-//****************************************************************************
 #include <Arduino.h>
-
-//****************************************************************************
-// @Defines
-//****************************************************************************
-#define BUFFER_LENGTH 32
-// WIRE_HAS_END means Wire has end()
-#define WIRE_HAS_END 1
```
LinjingZhang, line 38: I would be careful when removing these defines... this sometimes affects third-party libraries:
https://github.com/search?q=%23ifdef+WIRE_HAS_END&type=code

libraries/Wire/src/Wire.cpp
```diff
@@ -59,192 +14,133 @@ TwoWire::TwoWire(XMC_I2C_t *conf) {
...
-                                               (uint32_t)XMC_I2C_CH_EVENT_ALTERNATIVE_RECEIVE));
+            XMC_I2C_CH_DisableEvent(XMC_I2C_config->channel,(uint32_t)((uint32_t)XMC_I2C_CH_EVENT_STANDARD_RECEIVE |(uint32_t)XMC_I2C_CH_EVENT_ALTERNATIVE_RECEIVE));
         }
 
         flush();
-
         /* Disable i2c tx fifo */
-        XMC_USIC_CH_TXFIFO_Configure(XMC_I2C_config->channel, 16, XMC_USIC_CH_FIFO_DISABLED,
-                                     (uint32_t)15);
+        XMC_USIC_CH_TXFIFO_Configure(XMC_I2C_config->channel, 16, XMC_USIC_CH_FIFO_DISABLED,(uint32_t)15);
```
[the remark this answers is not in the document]
dineshgit411: while debugging the code , its feel like annoying for me. so i have changed the format to easily understanding.
and i removed unnecessary comments.

libraries/Wire/src/Wire.h
jaenrig-ifx: If it was in the header file... yep, that could be considered part of the public API... But we are also refactoring that API with breaking changes. So some "breaking things" might be justified. 
In this particular case I cannot tell, if this seems a standard define, then yes, better keep it.

LinjingZhang:
https://github.com/Infineon/XMC-for-Arduino/actions/runs/16517787835

Good job! HIL passed. So I will approve.

PR #361, opened 2025-07-17
Core api update iic modify

dineshgit411:

Description
Implemented Ringbuffer, error handling in endTransmission() function.

Related Issue
Every 20 to 30 seconds getting one or two iteration fail during Master pingpong and slavepingpong test case running. Suspecting timeout mechanism.
image

Context
Tested with two XMC 4700 board which one is master and one is slave, used 4.7K ohm resistor 
and Tested with Mastersender  - slavereceiver and Master Receiver - Slave sender arduino examples this is also working.
image

this screenshot of single board pingpong test multiple wire pins
image

### DIFF
--- a/libraries/Wire/src/Wire.h
+++ b/libraries/Wire/src/Wire.h
@@ -51,7 +51,6 @@ private:
     void (*user_onReceive)(int);
     void OnRequestService(void);
     void OnReceiveService(uint8_t numBytes);
-    void resetBus();
 };
 
 extern TwoWire Wire;
--- a/libraries/Wire/src/Wire.h
+++ b/libraries/Wire/src/Wire.h
@@ -5,7 +5,7 @@
 #include "api/RingBuffer.h"
 #include "api/HardwareI2C.h"
 
-#define WIRE_COMMUNICATION_TIMEOUT 5000u
+#define WIRE_COMMUNICATION_TIMEOUT 1000u
 #define BUFFER_LENGTH 64
 
 class TwoWire : public arduino::HardwareI2C {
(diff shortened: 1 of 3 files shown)


LinjingZhang:
I see that you have added some return values. So does the return value returns 4 to indicate a timeout error?

dineshgit411:
> I see that you have added some return values. So does the return value returns 4 to indicate a timeout error?

yes its returns timeout error, i have tested multiple times whenever first time fails its return timeout error return value
and i need to change the return value from 4 to 5 as per I2C Arduino documentation (thanks juan)

libraries/Wire/src/Wire.h
```diff
@@ -51,7 +50,8 @@ class TwoWire : public arduino::HardwareI2C {
     void (*user_onRequest)(void);
     void (*user_onReceive)(int);
     void OnRequestService(void);
-    void OnReceiveService(void);
+    void OnReceiveService(uint8_t numBytes);
+    void resetBus();
```
jaenrig-ifx, line 54: This needs removal?

dineshgit411:
Removed resetBus() function,  for debugging i added before. this one no longer needed , After removed also tested with multiple times its passing.

jaenrig-ifx (commented):
Specifically the requestFrom() and endTransmission() are for me hard to judge.
As long as the functionality is sufficiently tested, I would now consider that (for now) a black box :D 

Something that I think would help is to have an Wire tests, which are a bit more sequential and step by step checking the API of Wire.
That would help to easier assess each function, modes and cases than: 
https://github.com/Infineon/arduino-core-tests/tree/main/src/corelibs/wire

dineshgit411:
https://jirard.intra.infineon.com/browse/DESMAKERS-4647 -For time error in pingpong test case

https://jirard.intra.infineon.com/browse/DESMAKERS-4648 -  I2C tests for all functionality

LinjingZhang (approved):
👍

PR #376, opened 2025-07-29
Core api tone compile

dineshgit411:

Description
Modified Tone implementation for compilation error in all the boards.

Related Issue
Test case is passed but it runs only one need to check.
Context
image
tested with 4700 board.


LinjingZhang:
> > Nice!
> > I did not check details of the TONE changes. Is now still using ccu for pwm generation and wiring_time for timing?
> 
> yes, because i have seen in comments upto 500HZ only it could work if we use toggle method. so i kept CCU for pwm generation , only changed timing method, reused from 3.x.x releases

Does this mean we need to adjust the number of NUM_TONE_PINS?

LinjingZhang (approved):

PR #394, opened 2025-09-12
Optimize hil order

LinjingZhang:

Description
Provide the information we need to review your PR. What problem does the pull request solve? "Bug fix" is not a good description.

Related Issue
If you opened an issue before creating the PR, point to it here.

Context
What do we need to know about your development environment, tools, target, and so on. Screenshots are always helpful if there is a UI element to this PR.

### DIFF
--- a/config/user.yml
+++ b/config/user.yml
@@ -13,18 +13,21 @@
 unit-test:
    - unit-test-digital-io-4700
    - unit-test-tone-4700
+   - unit-test-iic-pingpong-4700
    - unit-test-interrupts-4700
    - unit-test-random-4700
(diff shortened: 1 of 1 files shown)


LinjingZhang:
> Why this reorder?

We currently have 4 jobs running in parallel (which is something we may need to change later).
If we put tests on the same board together, it would be no different from sequential testing. So to speed things up, I changed the order and tested. 
For some reason, draft PRs won't trigger the workflow.

LinjingZhang:
close since https://github.com/Infineon/XMC-for-Arduino/pull/393

PR #399, opened 2025-09-22
libraries/OneWire: Updated onewire library.

dineshgit411:

Description
Updated lates Onewire Library from https://github.com/PaulStoffregen/OneWire.git 
Related Issue
Onewire Library not worked in previous changes, after updated the library issue resolved
Context
image
Tested with XMC4700 relax kit


jaenrig-ifx:
> 6) || defined(ARDUINO_ARCH_XMC) like this

Well, not sure if the ARCH_XMC requires its own block. I would add a different one, even if it is the same. 2 independent blocks that we can modify separately if required.
But the usage of the official external library could be tested as well or? 
If the official external library works, we do not need to add it to our core, or?

jaenrig-ifx:
> > > 6. || defined(ARDUINO_ARCH_XMC) like this
> > 
> > 
> > Well, not sure if the ARCH_XMC requires its own block. I would add a different one, even if it is the same. 2 independent blocks that we can modify separately if required. But the usage of the official external library could be tested as well or? If the official external library works, we do not need to add it to our core, or?
> 
> About the macro, after check the PR [PaulStoffregen/OneWire#157](https://github.com/PaulStoffregen/OneWire/pull/157) I realize the code block is as same as the default configuration. Given that this PR hasn't been merged yet untill now, perhaps we could share a single PR and just add the xmc part?
> 
> The oneWire test will be added soon, only if it passes, we will merge this PR. The source Dinesh will delete soon.

Both architecture have nothing to do with each other. So each one having its own block (even if it is the same) allows the two cores integration to be maintained independently. For the same reason, maybe easier to create a separate PR for the reviewer.

Another option would be to add a list of boards that use the default without warning... so we don´t need to create a ifdef block for those defaults, and only create the custom one if required.

PR #405, opened 2025-10-23
Core xmc14 2go bug

dineshgit411:

Description
If you press the reset button mcu not reset.

Related Issue
Reset is not working in XMC14 2Go board, due to macro missing in pins arduino.
Context
image
.


jaenrig-ifx (approved):
All good!
I see that the arduino-core-api is updated to the last master commit. 
It is fine because there are only 2 commits ahead, and it seems not code relevant. 
But the master does not necessarily needs to be 100% complete or functional. So better update to tagged releases like -> https://github.com/arduino/ArduinoCore-API/releases/tag/1.5.2

Issue #439, opened 2026-06-08
Missing Implementation for  __FlashStringHelper related String functions


MaxIfx:
**Prerequisites**

Can you reproduce the problem reliably?
Yes, undefined reference errors are thrown when trying to use the SimpleSerialShell.h library

Did you check current release notes for known issues?
- 

If this is not the latest release, have you checked newer releases? 
I've checked with the latest 3.5.5 Version

**Description**:

When registering new commands using the F() Macro using SimpleSerialShell.h, the compiler throws undefined reference errors related to the string library. Looking at the source code of WString.h and WString.cpp in the XMC Core it looks like the declaration for the following functions exist in the header but there is no actual implementation:

- String & String::copy(const __FlashStringHelper *pstr, unsigned int length)
- String &String::operator=(const __FlashStringHelper *str) 
- String::String(const __FlashStringHelper *pstr)

The compiler issues can be resolved by copying the implementation from the AVR Core, however I have not tested if the code works with this as of yet.

**Steps to Reproduce:**

1. Install SimpleSerialShell.h and try to configure custom functions

**Expected Result:**

No Compiler Error 😃 

**Actual Result:**

Compiler Error 🥲 

**Frequency (1/1, 1/50, 1/100):**

1/1

**Build/Commit:**

3.5.5

**Target:**

XMC1100 Boot Kit

**Host OS and Version:**

Windows

**Compiler:**

Include the version!

**Environment**

Provide the context surrounding this problem. Perhaps an IDE, a debug probe, an RTOS, etc. (and versions)

jaenrig-ifx:
Hi @MaxIfx,

Did you try the latest version 4.x? ->https://github.com/Infineon/XMC-for-Arduino/releases/tag/4.1.1
These functions should be supported in that version.
Any reasons to stay with 3.x?

MaxIfx:
Hi @jaenrig-ifx,

3.5.5 is the latest Version that is shown in my Board Manager so I guess the paths are not up to date in the installation. Thanks for updating the Files!

jaenrig-ifx:
> Hi [@jaenrig-ifx](https://github.com/jaenrig-ifx),
> 
> 3.5.5 is the latest Version that is shown in my Board Manager so I guess the paths are not up to date in the installation. Thanks for updating the Files!

You'll have to update the _Additional Boards Manager URL_ with:

```
[...]

PR #440, opened 2026-06-10
tools/ci.sh: Fix commit check base branch for 3.x

ederjc:
## Problem

Two CI checks were always failing for feature branches based on `master-2.x-3.x`:

**1. Commit message check (`commit_formatting.yml`)**
`ci_commit_formatting_run` in `tools/ci.sh` compared against `upstream/master`. Since `master-2.x-3.x` diverged far from `master`, this caused `verifygitlog.py` to check all 700+ historical commits — many of which were merged via GitHub with `noreply` email addresses — making the check always fail regardless of the new commit's quality.

**2. HIL unit tests (`hil-unity-checks.yml`)**
The HIL workflow requires a self-hosted runner with physical XMC hardware boards connected. This infrastructure is not configured for the `master-2.x-3.x` branch, so the workflow fails on every push.

## Fix

- **`tools/ci.sh`**: Changed base reference from `upstream/master` to `upstream/master-2.x-3.x`, and added `--ignore-rebase` to match the `master` branch behaviour. Only commits added by a feature branch are now checked.
- **`.github/workflows/hil-unity-checks.yml`**: Removed from this branch until the HIL self-hosted runner infrastructure is set up to support `master-2.x-3.x`.

## Related

Discovered while working on [#439](https://github.com/Infineon/XMC-for-Arduino/issues/439) — missing `__FlashStringHelper` String implementations in `WString.cpp`.


LinjingZhang (approved):
👍 nice

